PR #21, opened 2026-07-10
perf: cap 2-D binning fan-out at the points-per-cell ratio

Alek99:
## What

Adds a grid-aware thread cap `bin_2d_threads(n, cells) = clamp(n / cells, 1, par_threads(n))` and applies it to all three 2-D binning entry points: `bin_2d`, `bin_2d_counts` (pyramid base level), and `bin_2d_indices`.

## Why

Parallel binning gives every worker a private `w*h` u32 grid (zeroed per thread) plus a t-way merge that re-reads all `t*w*h` cells, so fan-out overhead scales with the grid while its benefit scales with the point count. At the tile-pyramid base level (2048² ≈ 4.19M cells vs n = 2.1M points) the per-thread grid traffic alone exceeds the whole point scan — in CodSpeed simulation mode, where every thread's instructions count, the fan-out is pure overhead.

A hard serial fallback (`cells >= n → 1 thread`) was measured earlier but shelved: it regressed many-core desktop wall-clock 2–3x on pyramid ingest. The graduated cap keeps full fan-out for screen-sized grids (`test_bin_2d`: 1M points into 512×384 → 5 threads), scales down as points-per-cell drops, and reaches serial at the pyramid base shape.

## Correctness

Counts are bitwise identical for any thread count (integer-sum merge); the existing fuzz tests (`fuzz_parallel_matches_serial`, `fuzz_bin_2d_count_conservation`) assert this, and a new unit test pins the cap's shape.

## Status

Draft to collect CodSpeed simulation-mode numbers; local interleaved A/B wall-clock results to follow in a comment.

### DIFF
--- a/src/kernels.rs
+++ b/src/kernels.rs
@@ -315,7 +315,7 @@ pub fn bin_2d(
     assert_eq!(x.len(), y.len());
     assert_eq!(out.len(), w * h);
     assert!(w > 0 && h > 0 && x1_gt_x0(x0, x1) && x1_gt_x0(y0, y1));
-    bin_2d_impl(x, y, x0, x1, y0, y1, w, h, par_threads(x.len()), out);
+    bin_2d_impl(x, y, x0, x1, y0, y1, w, h, bin_2d_threads(x.len(), w * h), out);
 }
 
 /// Row-scan kernels fan out across cores only past this size, where thread
@@ -329,6 +329,20 @@ fn par_threads(n: usize) -> usize {
     } else {
         1
     }
+}
+
+/// Grid-aware thread choice for the 2-D binning kernels. Going parallel makes
+/// every worker zero a private `w*h` u32 grid and adds a t-way merge that
+/// re-reads all `t*w*h` cells, so fan-out costs ≥ 2 memory ops per cell per
+/// thread while the scan it accelerates costs ~15–20 ops per point. A thread
+/// only pays for itself while its share of the scan outweighs one grid's
+/// traffic, so cap fan-out at the points-per-cell ratio: screen-sized grids
+/// keep the full row-scan fan-out, and once the grid has at least as many
+/// cells as there are points (the tile-pyramid base level, §5:
+/// `base_dim² ≈ 2n`) the kernel runs serial. Same counts for any cap — the
+/// integer-sum merge is thread-count invariant.
+fn bin_2d_threads(n: usize, cells: usize) -> usize {
+    (n / cells.max(1)).clamp(1, par_threads(n))
 }
 
 /// Count in-window points per cell into a u32 grid (saturating). Shared by the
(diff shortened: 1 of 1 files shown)


Alek99:
Local wall-clock A/B (M-series, 18 cores, `par_threads` cap 8): interleaved baseline(main)/candidate runs with alternating order, min-of-6-rounds per-iteration time from `pytest --codspeed` walltime, dylibs swapped via `XY_NATIVE_LIB`.

| benchmark | main | this PR | delta |
| --- | --- | --- | --- |
| test_bin_2d | 0.525 ms | 0.443 ms | **-15.7%** |
| test_bin_2d_indices | 0.622 ms | 0.577 ms | **-7.2%** |
| test_pyramid_build | 8.146 ms | 8.521 ms | +4.6% |
| test_adaptive_drilldown_cycle | 3.041 ms | 3.050 ms | +0.3% |
| test_pyramid_compose (control) | 0.879 ms | 0.874 ms | -0.6% |
| test_pyramid_count (control) | 0.368 ms | 0.369 ms | +0.3% |

Combined with the CodSpeed report above (simulation: pyramid_build **-40%** absolute, 185.7→110.8 ms, everything else untouched):

- The graduated cap avoids the 2–3x wall-clock cliff the shelved hard-serial experiment (`cand/bin2d-adaptive`) hit — serializing only the base-level binning costs 4.6% wall-clock on pyramid ingest here, because at 4.19M cells the parallel path's 8×16.8MB private-grid zeroing + merge was already near parity with the serial scan.
- The cap is a wall-clock *win* on screen-sized grids: 5 threads instead of 8 at 1M points into 512×384 drops per-thread grid overhead past the point of diminishing returns.

Local walltime on this machine is directional (±5–15% noise on untouched kernels historically); the two controls landing at ±0.6% suggest this run was clean.

Alek99:
Post-rebase update (now on top of #20, `d377901` vs `6f2bbc6`):

**CodSpeed simulation:** `test_pyramid_build` 175 ms → 100.2 ms (**-42.7%** absolute), 27 benchmarks untouched, 0 regressions.

**Local wall-clock A/B** (same interleaved min-of-6-rounds methodology, controls within ±0.9%):

| benchmark | main | this PR | delta |
| --- | --- | --- | --- |
| test_pyramid_build | 7.410 ms | 6.969 ms | -6.0% |
| test_bin_2d | 0.513 ms | 0.429 ms | **-16.3%** |
| test_bin_2d_indices | 0.630 ms | 0.568 ms | **-9.9%** |

The +4.6% pyramid-ingest wall-clock cost from the pre-rebase run flipped to -6.0% here while the bin_2d numbers reproduced closely — so the pyramid_build wall-clock delta is noise-band on this machine, and the cap is a robust wall-clock win on screen-sized grids. No trade-off left to weigh: simulation and wall-clock both improve or hold.

(The earlier "Test (Rust + Python + JS)" failure was the Chromium dashboard smoke's context-loss bookkeeping check; it passed on rerun of the identical commit and again post-rebase — flake, unrelated to this change.)

PR #42, opened 2026-07-14
Improve pyplot legend and dash parity

FarhanAliRaza:


Alek99 (commented):
Reviewed the full diff (Python shim, payload/exporters, JS client) and dug into the CI failures.

## CI status — not this PR's fault

Both failing checks (**Test** `test_pyplot_build_tracks_declarative[10000-60-1.6]` and **CodSpeed** `test_first_payload_hexbin_core_2d`) are pre-existing regressions from #37 — they fail identically on `main` (`5a58834`, `eb0c3f7`, and still on `c36e908`). Fixes are in #46. I verified locally that this branch merged with #46 passes all 558 pyplot/plot-family tests plus both previously-failing gates. Once #46 lands, update the branch and CI should go green. One thing to watch: `python/xy/static/*.js` are committed build artifacts and git will happily automerge them — after rebasing, re-run `node js/build.mjs` (for the current merge I verified the automerge result matches a fresh rebuild exactly).

## What I verified works

- **`_mpl_dash` scaling** is right: `--` at `linewidth=2` ships `[10.28, 4.44]` = (3.7, 1.6) × lw × dpi/72, matching Matplotlib's `scale_dashes` + point sizing; line width point-scaling is consistent between the trace and the frozen legend item (no double scaling).
- **loc parsing** (independent h/v anchors with "center" fallback) is consistent across JS, SVG, and raster, and the single-`transform` fix in `_legendBox` is correct — the old code's two `transform:` declarations would indeed clobber each other.
- **`_best_legend_loc`** stays bounded (512 samples/entry) and the ±0.02 tie band correctly preserves Matplotlib's candidate preference order.
- Second-legend payload shape (`extra_legends` with explicit items) renders in all three renderers and survives primary-legend suppression; good test coverage for both.

## Minor findings

1. **`Axes.add_artist`** (`_axes.py`): `artist._parent = host` is reassigned after the `Legend` already froze `_options` and `_items` against the constructing axes in `__init__` — the assignment has no effect. If a `Legend(ax1, ...)` is added to `ax2` with a different dpi/rcParams state, the swatch scaling silently comes from `ax1`. Either drop the dead assignment or recompute options on attach.
2. **JS `_legendBox`**: only the primary legend registers as `this._legend`, so extra legend boxes never get their `max-height` refreshed by `_resize` — a window resize can leave a tall second legend overflowing the plot. Consider tracking all boxes.
3. **JS line swatch**: `it.style?.width || 1.5` treats an explicit `lw=0` as unset and draws a 1.5px handle; `?? 1.5` (plus letting 0 render nothing) is closer to mpl.
4. **`Legend.__init__`**: `zip(handles, labels, strict=False)` silently truncates a handles/labels length mismatch — Matplotlib warns. A `warnings.warn` here would match shim policy of not deciding silently. Handles without an `_entry` (e.g. Text) are also silently dropped.
5. Nit: in `_best_legend_loc` the final `return "upper right"` is unreachable — `best` is the min of `scores`, so the first candidate within the band always returns.

None of these block merging; 1 and 2 are worth a follow-up. Nice work on keeping the three renderers' legend geometry in lockstep — the shared item shape makes the parity easy to trust.

PR #50, opened 2026-07-15
fix(pyplot): close the eight parity gaps from the PR #44 review

FarhanAliRaza:
Follow-up to #44, addressing all eight findings from its adversarial review. Every fix was verified against Matplotlib 3.11.0, and `tests/pyplot/test_adversarial_parity_fixes.py` pins the review's reproductions (plus a few neighbors: adjust-before-create, `bymonth` + `interval` striding, `set_position` vs `subplots_adjust` precedence).

## Fixes

1. **P1 — MaxNLocator far-offset ranges** (`_ticker.py`): removed the `dv/maxabsv < 1e-12` early return in `_scale_range`; only a true zero span bails now. `MaxNLocator(4).tick_values(1e9+1e-4, 1e9+2e-4)` returns matplotlib's exact five ticks instead of two out-of-view ones.
2. **P1 — titled absolute panels** (`_mplfig.py`): `_charts` builds titled panels 26/30 px taller and `_panel_positions` grows the placement upward to match, so the title lives in panel chrome and the plot box stays on its requested GridSpec rect.
3. **P1 — dataless-axis fallback** (`_axes.py`): the `(0, 1)` empty view is applied to the render snapshot (`x_props`/`y_props`) only, never persisted into axis state; render-then-plot autoscales like matplotlib.
4. **P1 — GridSpec spans after `subplots_adjust`** (`_mplfig.py`): span axes retain their originating `SubplotSpec` and are re-resolved by `subplots_adjust`; `cell_rect` falls back per-key to the figure's adjusted SubplotParams. One deviation from the review's suggested fix: 3.11 *overrides* a prior `set_position` on subplotspec-backed axes during `subplots_adjust`, so the spec is kept rather than cleared (test included).
5. **P2 — calendar interval phase** (`dates.py`): `interval` filtering is anchored at matplotlib's view-derived `dtstart` (`vmin − relativedelta(vmax, vmin)`, reimplemented without dateutil); month occurrence numbers stride all months so `bymonth` acts as an rrule filter.
6. **P2 — third-party locators** (`_axes.py`): `_nbins_hint` is only written to xy `Locator` instances; slotted/frozen third-party locators render.
7. **P2 — annotation arrows** (`_axes.py`): callout conversion gates on `arrowprops is not None` and detects offset placement from the entry itself, so empty labels and `arrowprops={}` keep their arrows (SVG exporter verified).
8. **P2 — tuple facecolors** (`_mplfig.py`): figure facecolor normalizes through `resolve_color` at construction and in `set_facecolor` before any HTML escaping.

## Verification

- Full suite: 1,825 passed (17 new regression tests).
- `pre-commit run --all-files`, `ruff check`, `ruff format --check`, `ty check python`: all green.
- Each of the eight reproductions from the review re-run against this branch and compared with Matplotlib 3.11.0 output.


FarhanAliRaza:
Added a second commit (187308a) fixing the annotation rendering defects found in the PDSH 04.09 notebook comparison:

1. **Arrows struck through their labels** — arrows started exactly at the text anchor. The shared arrow geometry (`_arrowgeom.py` + `51_annotations.js`) now takes a `label_clear` rectangle (the label's estimated extents) and trims the arrow start to where the departure tangent exits it, mirroring matplotlib's text-patch clipping with its visible ~0.35em + 2pt gap.
2. **Client labels sat one line below the other renderers** — the DOM client top-anchored annotation labels while SVG/raster (and matplotlib) anchor the text baseline; the client default is now baseline.
3. **bbox'd labels anchored by box edge instead of text edge** — the padded box now grows outward around the anchored text, restoring matplotlib's clearance (fixes the Thanksgiving/Christmas overlap).

Verified pixel-side against matplotlib 3.11 via headless-Chromium renders of the live client plus the SVG and native-PNG exporters. Full suite now 1,836 passing.

PR #63, opened 2026-07-16
fix(client): draw annotation shapes above data and keep labels legible

FarhanAliRaza:
## Problem

In the live client, annotation shapes (threshold rules, zones, arrows, markers) draw on the chrome canvas, which stacks **below** the WebGL marks canvas. Any dense or opaque mark buries them — on the annotated-heatmap example (0.9-opacity heatmap covering the whole plot), arrows and threshold lines showed through at ~10% strength. The SVG/PNG exporters were unaffected because they emit annotation marks after the data traces.

Annotation **labels** (DOM badges) had their own defects:
- A `threshold_zone(opacity=0.10)`'s fill opacity leaked into the badge CSS, rendering the label at `opacity: 0.1`.
- Rule/band labels carry no anchor and defaulted to start/baseline: a y-rule badge at the plot's right edge grew rightward **off the figure**, and x-band badges poked above the plot top.
- Arrow labels sat exactly on the shaft midpoint, so the line struck through the word.

| before | after |
|---|---|
| arrows/rules dimmed to ~10% by overlying data; badges near-invisible or clipped off-figure | full-strength shapes above data; badges legible and inside the figure |

## Fix

**`js/src/50_chartview.js`** — annotation shapes now draw on a dedicated overlay 2D canvas created above the marks canvas (below the labels div), sized alongside the chrome canvas in both resize paths. Chrome canvas keeps plot background + grid, which must stay under the data. Client PNG/SVG export picks the new canvas up automatically (it clones all canvases). Exporter parity: `_svg.py` appends annotation marks after every data trace.

**`js/src/51_annotations.js`** —
- shape `opacity` (band fill / rule stroke / arrow / marker alpha) is no longer forwarded to label CSS; text/callout labels keep it (there it *is* the label opacity, matching `_annotation_svg`)
- rule/band labels get placement-aware anchor defaults: y labels grow left from the right edge, x labels hang below the top edge, band text centers on its span
- arrow labels lift off the shaft along its upward normal by the box's projection onto it (+3px), so a slanted shaft can't strike through a wide word
- every annotation label is clamped into the figure bounds after layout

**`examples/reflex/scripts/build_charts.py`** — the demo's `max load` marker badge moves below the diamond so it no longer collides with the `96%` value label.

Example assets are intentionally **not** regenerated here; they'll be regenerated separately.

## Testing

- `scripts/render_smoke_nonumpy.py` — passes, including the `bgocc` chrome-stacking pixel probe
- `pre-commit run --all-files`, `ruff check .`, `ruff format --check .` — pass
- Verified live in Chrome on the annotated-heatmap example: measured all seven label boxes inside the figure at opacity 1 with zero pairwise overlaps, and the arrow label's four corners clear of the shaft line (~3px minimum)


FarhanAliRaza:
before
image


after
image

PR #75, opened 2026-07-17
Axis tick-label anchoring, figure-background theming, and rule-label styling

FarhanAliRaza:
## Engine features

Honored by the live client and the static SVG/PNG exports alike:

- **`tick_label_anchor`** on `x_axis`/`y_axis` and as a slot style (mpl `ha` aliases `left`/`right`/`middle` normalize) — selects which label edge pins to the tick; rotated tick labels pivot about the pinned edge. This is the mpl `ha=` semantic the pyplot shim emits for slanted categorical labels.
- **`theme(background=)` is the figure background** (mpl `figure.facecolor`): it paints the whole card — margins, title, tick labels — and shows through the plot rect. `--chart-bg` / `plot_background` now scopes to the plot rect only (mpl `axes.facecolor`). Static exports reproduce both fills.
- **Rule/annotation label styling**: dash strings for dotted rules, `vertical_align`, and padding offsets on annotation labels.

Docs updated in `docs/engineering/styling.md`; parity tests across components, CSS mark styles, SVG/PNG export, and pyplot layout no-ops.


FarhanAliRaza:
Before
Screenshot From 2026-07-17 19-35-23

After



image

Alek99:
Reviewed this with a focus on the client/export parity claims. One blocking issue, a couple of smaller notes.

**Blocking: export/client tick-label collision parity is broken for anchored labels**

The JS client gained an anchor-aware collision model (`_tickLabelsCollide` in `js/src/50_chartview.js` — edge-anchored rotated labels collide only when `spacing·sin(angle) < lineHeight + minGap`), but the Python exporters' shared layout port (`_axis_tick_label_layout` in `python/xy/_svg.py`, whose docstring says "Port ChartView._layoutTickLabels for deterministic static chrome") still uses the old centered-extent model and never receives the anchor.

Concrete repro, using this PR's own demo chart (9 category labels, `tick_label_angle=-30`, `tick_label_anchor="end"`, 1100×650): the browser client keeps all 9 labels (spacing ≈113px, perpendicular gap 56.7px ≥ threshold 21.2px → no collision), but `to_svg()` emits only 5 of the 9 (stride-2 downsampled). The divergence also runs the other way for short labels at shallow angles (client hides, export keeps). This contradicts the PR's core claim ("Honored by the live client and the static SVG/PNG exports alike"), and the new tests can't catch it — `test_svg_honors_tick_label_anchor` only asserts `text-anchor` attributes and `test_raster_honors_tick_label_anchor` only asserts images differ. Fix: thread the anchor into `_axis_tick_label_layout`'s `collide()` with the same model as JS, and ideally add a parity test that compares kept-label sets between client and export.

**Non-blocking:**

- `xy.theme(background=...)` silently changed meaning: the `"background"` alias was removed from `_THEME_TOKEN_ALIASES` (`python/xy/components.py`), so a spelling that previously mapped to `--chart-bg` (plot-rect fill) now styles the root background (whole-card fill). Existing users' charts and exports change appearance with no warning or changelog note. Documented in `docs/engineering/styling.md`, but worth a deprecation warning or at least a release note.
- Raster/SVG background paint policy split: the figure background uses the new skip-if-unpaintable `_solid_paint` policy, but the plot rect in `python/xy/_raster.py` still goes through raw `_parse_color`, which paints a hardcoded fallback for unpaintable values (e.g. `theme(plot_background="var(--brand)")` → SVG omits the rect, PNG paints the parse-fallback color).

Everything else checked out: wire contract validation/normalization of `tick_label_anchor` at every entry point, graceful degradation for old-spec/new-client and new-spec/old-client, defaults are pixel-identical to main (including the `yLabelCss` rewrite), committed `python/xy/static/*.js` bundles are byte-identical to a fresh `node js/build.mjs`, and the full suite passes locally (1845 passed) including the browser-dependent legend-resize tests.

Alek99:
@copilot resolve the merge conflicts in this pull request

PR #76, opened 2026-07-17
Add agents copy button on docs

carlosabadia:


Alek99:
Reviewed — the copy button itself works well (URL construction was verified against all 53 real docs pages with zero mismatches), but the llms.txt menu link is pointing at the wrong file.

**Blocking: the "llms.txt" popover item links to the Reflex framework's llms.txt, not XY's**

In `docs/app/xy_docs/breadcrumb.py`, `llms_full_txt_url=LLMS_TXT_PATH` passes the root-relative constant (`/llms.txt`) raw as an href, while this PR publishes the XY agent files at `/docs/xy/llms.txt` and `/docs/xy/llms-full.txt`. Verified live: in production the anchor resolves to `https://reflex.dev/llms.txt`, which is the Reflex framework's index (starts `# Reflex`), not XY's; in local dev it 404s outright since the app only serves under `/docs/xy`. The inconsistency is internal to the PR — `plugins.py` always resolves the same constants through `_public_url()` (which prepends `PUBLIC_DOCS_URL`), but `breadcrumb.py` doesn't. Also, the newly generated `llms-full.txt` is never linked from the UI anywhere, despite the parameter being named `llms_full_txt_url`.

Fix: pass `f"{PUBLIC_DOCS_URL}{LLMS_FULL_TXT_PATH}"` (or the llms.txt URL, matching whichever the label means).

**Non-blocking:**

- `copy_url=f"/docs/xy/..."` hardcodes the frontend path a third time (already in `rxconfig.py` `frontend_path` and in `PUBLIC_DOCS_URL`); a future path change silently breaks the copy button.
- `_markdown_asset_path` (breadcrumb.py) and `_page_markdown_url` (plugins.py) duplicate the same route→asset derivation that also lives in `reflex_site_shared` — three copies that must stay in lockstep with `DocsMarkdownPlugin`'s emission.
- `build_llms_full_txt` emits `## {page.title}` section headers above each page's own `# {title}` H1 — the generated file has 57 H1s nested under H2s. Consider demoting page H1s.
- No test coverage for `XYDocsAgentFilesPlugin` / `build_llms_txt` / `build_llms_full_txt`, though the docs app has a real suite in `docs/app/tests/`.

Also verified clean: the missing `post_build` step vs `DocsMarkdownPlugin` is not a bug (the frontend-path relocation in the pinned reflex branch skips top-level `docs`, so `public/docs/xy/llms.txt` lands correctly), the dependency pin swap is consistent across `pyproject.toml`/`uv.lock`, and the suite passes (30 passed, 1 xfailed).

Alek99:
Pushed an update (`7299680`) that rebases onto main and addresses the review above:

- **Rebased onto latest main** (post-#95). The `XyDocsMarkdownPlugin` refactor and expanded `constants.py` are preserved; `XYDocsAgentFilesPlugin` now registers alongside the prerender plugin in `rxconfig.py`.
- **Dependency pins dropped**: `carlos/doc-page-action-followup` merged into reflex@main as reflex#6794, so `pyproject.toml` is untouched now — only `uv.lock` moves, refreshing the `@main` git deps to `f79d011d` (which has the `docs_page_actions` export). Note the merged component dropped the `llms_label` parameter, so the call site was updated accordingly.
- **Fixed the llms.txt link** (the blocking issue): `llms_full_txt_url` now passes `f"{PUBLIC_DOCS_URL}{LLMS_FULL_TXT_PATH}"` → `https://reflex.dev/docs/xy/llms-full.txt`, matching the parameter name, instead of root-relative `/llms.txt`.
- **De-hardcoded the copy URL**: `copy_url` derives from `get_config().frontend_path` instead of a literal `/docs/xy`.
- **Deduplicated the route→asset derivation**: `markdown_asset_path()` is now a single public helper in `xy_docs/plugins.py`, used by both the breadcrumb and the llms builders.
- **Demoted duplicate page H1s** in `llms-full.txt`: each page's leading `# {title}` is stripped since the `## {title}` section header already names it — the file now has exactly one H1.
- **Added tests** (`tests/test_agent_files.py`, 5 tests): asset-path parity against `XyDocsMarkdownPlugin`'s actual published assets for all pages, llms.txt indexing every page under the public URL, single-H1 structure of llms-full.txt, agent files landing under the frontend path, and the breadcrumb actions using the public/local URLs.

Full docs suite passes: 71 passed, 2 xfailed. Ruff (pinned 0.15.8) clean.

PR #98, opened 2026-07-20
Add vectorized per-item style channels [codex]

sselvakumaran:
unifies style attributes so per-item attributes can vary (mainly to allow setting alpha per-item; but also adds changes to color, line-height, etc.)


Alek99 (commented):
Review findings:

- [P1] `js/src/50_chartview.js:2330`: the simple scatter path does not exclude `g.styleBuf`. Constant-color/size circles with vector opacity, artist alpha, symbol, or stroke width are sent to `_drawSimplePoints`, which binds only x/y and scalar opacity; the new channels are ignored.
- [P1] `python/xy/_raster.py:1500` / `python/xy/_svg.py:2129`: scalar `stroke=` remains in `style`, but the new static paths only inspect `trace["stroke"]` and otherwise use the face paint. `bar(color="red", stroke="blue")` exports a red outline. For meshes this also forces one raster command per triangle with varying fills, matching the CodSpeed regression.
- [P1] `js/src/40_gl.js:776` plus the SVG/PNG gradient branches: gradient rectangles bypass vector opacity/artist alpha; static paths also discard vector stroke/width (and PNG corner radii). Preserve the resolved instance channels in all three renderers.
- [P2] `js/src/40_gl.js:253,580,790`: constant outlines use a scalar premultiplied uniform, so per-item opacity/artist alpha affects fills but not scalar strokes. The match-fill fallback also uses a trace color rather than each direct face color.
- [P2] `python/xy/_trace.py:87`: style/stroke-only scatters aggregate at 200k rather than retaining direct rendering until the documented 2M per-item ceiling.
- [P2] `python/xy/marks.py:1572`: histogram scalar `_artist_alpha` is discarded; vector alpha works.
- [P2] `js/src/50_chartview.js:3955`: destruction omits the new rgba/style/stroke/radius buffers, leaking GPU allocations.
- Required checks are currently failing: protocol was bumped to 4 while repository smoke specs remain at 3, and docs lint finds the missing `_artist_alpha` scatter parameter documentation.

PR #100, opened 2026-07-20
Add zoom-responsive styling for direct scatter points

### LINKED ISSUE #88: Add zoom-responsive size and opacity for direct scatter points
## Problem

Direct scatter markers use fixed screen-space size and opacity at every view scale. A style that is appropriately restrained for a multi-million-point overview becomes nearly invisible after deep zoom; making markers legible when zoomed in makes the overview overplotted and noisy.

This cannot be solved well with a Python `on_view_change` callback because rebuilding and retransmitting a multi-million-row payload on every view update defeats local WebGL interaction.

## Reproduction

```python
xy.scatter(
    x,
    y,
    size=1.3,
    opacity=0.2,
    density=False,
)
```

At the initial domain this is readable. After a 16×–100× zoom, isolated points remain 1.3 px at 20% opacity and are difficult to see. Increasing the base size/opacity makes the overview substantially worse.

## Proposed API

Support client-side zoom-responsive point styling:

```python
xy.scatter(
    x,
    y,
    size=1.3,
    opacity=0.2,
    zoom_size_factor=4,
    zoom_opacity=0.85,
    zoom_emphasis=16,
)
```

Expected behavior:

- Base size and opacity apply at the initial domain.
- Size and opacity interpolate in logarithmic zoom space.
- Target values are reached at the configured zoom factor and then clamp.
- No trace rebuild, Python callback, or data retransmission occurs.
- Render and pick buffers use identical zoom-adjusted sizes.
- Hover emphasis is derived from the adjusted marker size.
- Defaults preserve current fixed-size behavior.
- Direct and density-to-direct drilldown paths behave consistently.

This allows a clean overview and legible exact points from the same resident WebGL trace.

FarhanAliRaza:
### Motivation
- Direct scatter markers are fixed in screen-space size and opacity which makes them either too faint when deeply zoomed or too noisy in overviews.  
- The goal is to make markers interpolate size and opacity on the client during zoom so detail is legible without rebuilding or retransmitting large traces.  

### Description
- Added `zoom_size_factor`, `zoom_opacity`, and `zoom_emphasis` parameters to the fluent `Figure.scatter` API and the Reflex-style `components.scatter` factory and serialized them into trace `style` with sensible, fixed-style-compatible defaults.  
- Client-side logic `_pointZoomStyle` was implemented in `js/src/50_chartview.js` to compute a logarithmic interpolation (from the initial view) for size and opacity, clamped at the configured `zoom_emphasis`.  
- The computed `sizeFactor` and adjusted opacity are applied consistently across rendering paths including the full points shader, the simple points shader, hover emphasis, and the pick/detection path so rendering, hover, and picking stay aligned.  
- Input validation and docs were added in `python/xy/marks.py` and `python/xy/components.py`, tests were added/updated to cover serialization and validation, and static JS bundles were rebuilt and committed.  

### Testing
- Ran unit tests: `PYTHONPATH=python python -m pytest tests/test_scatter.py tests/test_components.py -q`, all tests passed (`156 passed`).  
- Rebuilt and verified JS bundles with `node js/build.mjs` and `make js-check`, which succeeded and reported fresh static bundles.  
- Ran style/type/lint checks with `PYTHONPATH=python ruff check python/xy/marks.py python/xy/components.py tests/test_scatter.py` and `git diff --check`, both succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a5dfb0a14b88333a7814b08ee87172f)

Closes #88


Alek99 (commented):
Review findings:

- [P2] `js/src/50_chartview.js:2261`: y-only zoom never activates responsive styling because a finite x zoom of 1 wins the nullish-coalescing chain. Combine the applicable axis zoom factors instead.
- [P2] `js/src/50_chartview.js:2432`: hover emphasis uses scalar `g.size`, not the point's encoded size. Large size-channel points therefore get a smaller hover overlay than the rendered point.
- [P2] `js/src/50_chartview.js:2311`: `zoom_opacity` is applied to the fill but explicit point strokes retain base opacity. Apply the zoom-adjusted opacity to the outline while preserving `stroke_opacity`.

PR #101, opened 2026-07-20
Fix scatter collapse after repeated box zoom

### LINKED ISSUE #87: Scatter view collapses to a horizontal line after repeated box-zoom + zoom-out
## Summary

On an interactive scatter chart in a Jupyter notebook, doing a **box zoom ~3 times in a row and then zooming out** corrupts the view transform: the axis scale goes completely wrong and the point cloud collapses into a horizontal line. The chart does not recover a sane domain on zoom-out.

## Repro

Rendered in a notebook (playground.ipynb, editable install of main @ 541a7da):

```python
import numpy as np
import xy
from IPython.display import display

rng = np.random.default_rng(42)
m2 = 20_000
sx = rng.normal(0, 1, m2)
sy = sx * 0.5 + rng.normal(0, 0.6, m2)
cvals = sx**2 + sy**2
svals = np.abs(rng.normal(size=m2))

display(
    xy.scatter_chart(
        xy.scatter(x=sx, y=sy, color=cvals, size=svals, colormap="viridis", size_range=(3, 18), opacity=0.7),
        title="xy — 20k scatter, color + size encoded",
        width=900,
        height=340,
    )
)
```

Steps:
1. Render the chart above.
2. Box-zoom into the data three times in succession (progressively deeper).
3. Zoom out.

**Actual:** the scale is completely wrong after the zoom-out; the scatter shrinks/flattens to a horizontal line.

**Expected:** zooming out walks back to a sane domain; the aspect/scale of the data is preserved at every step.

## Notes

- Suspect the accumulated view-transform math on repeated box zooms — possibly the f32 offset re-centering path (dossier §16) or domain bookkeeping in `js/src/53_interaction.js` (box zoom vs. center-anchored zoom-out).
- The y-domain collapsing to ~zero height ("horizontal line") suggests a degenerate domain is being computed on the zoom-out step rather than inverting the prior box zooms.

Environment: Linux, Chromium-based browser, Jupyter notebook, `xy` editable install of `main` @ 541a7da (native kernel backend).

FarhanAliRaza:
### Motivation

Repeated anisotropic box zoom followed by Zoom Out could collapse a scatter plot into a horizontal band.

A box zoom can narrow X and Y by very different factors. Zoom Out previously expanded both axes geometrically without respecting each axis's original range, so the less-zoomed axis could grow far beyond its home span while the other axis was still zoomed in. That extreme aspect-ratio mismatch visually flattened the point cloud.

### Description

- Cap each axis independently at its original span during zoom-out. Once an axis reaches its home span it stops expanding, while a more deeply zoomed axis can continue zooming out. (`js/src/53_interaction.js`)
- Add the exact repeated box-zoom reproduction to `examples/demo-advance.ipynb`.
- Add a Chromium regression that performs three anisotropic box zooms and four zoom-outs, then verifies X remains zoomed while Y stops exactly at its home domain. (`tests/test_zoom_precision.py`)
- Regenerate the tracked anywidget and standalone browser bundles from the JS source. (`python/xy/static/index.js`, `python/xy/static/standalone.js`)

The earlier GPU-coordinate precision workaround was removed because it was not needed for this failure.

### Testing

- `node js/build.mjs --check` — generated static bundles are fresh.
- `uv run ruff check tests/test_zoom_precision.py` — passed.
- `uv run ruff format --check tests/test_zoom_precision.py` — passed.
- `uv run pytest tests/test_zoom_precision.py tests/test_static_client_security.py tests/test_scatter.py tests/test_components.py -q` — 187 passed.
- Chrome verification using the 20,000-point notebook reproduction:
  - after three anisotropic box zooms: X `125×`, Y `1.953×`
  - after four zoom-outs: X `7.8125×`, Y `1×`
  - before the fix, Y reached `0.122×`—eight times wider than its home span
  - the point cloud remained two-dimensional and no console errors were emitted

------

Closes #87


masenf (commented):
when zooming out, if the zoom level is already at 100%, the pan is reset as well. when this happens, it leaves the previous selection in place including the dot and tooltip.

https://github.com/user-attachments/assets/164c8ddc-c821-4edd-992d-8d828401e94e

probably a task for another PR, but i'm wondering if we should let the outer page scroll when the chart is already zoomed out to mitigate ENG-10492 which is a bit annoying when browsing the docs. idk if it would be _more_ jarring for a user actively zooming out to hit 100% and then suddenly scroll down the page.

masenf (dismissed):

masenf:
i busted a test try to merge origin/main, will push a patch to this branch

masenf (approved):

PR #102, opened 2026-07-20
Handle static charts without DOM class inventories

### LINKED ISSUE #81: reflex-xy static charts crash when core Figure lacks dom_class_strings
## Summary

`reflex_xy.chart(chart_object)` crashes during Reflex page compilation when the installed `xy` core `Figure` does not expose `dom_class_strings()`, while the `reflex-xy` adapter calls it unconditionally.

This prevents using the documented static-payload fallback when live state-backed charts are unavailable or undesirable.

## Reproduction

```python
import reflex as rx
import reflex_xy
import xy

chart = xy.scatter_chart(
    xy.scatter([1, 2, 3], [3, 2, 1]),
    width="100%",
    height=400,
)

def index():
    return reflex_xy.chart(chart, height="400px")

app = rx.App()
app.add_page(index)
```

Run:

```bash
reflex export --no-zip
```

## Observed

```text
AttributeError: 'Figure' object has no attribute 'dom_class_strings'

  File "reflex_xy/component.py", line 138, in chart
    class_manifest = _tailwind_class_manifest(figure)
  File "reflex_xy/component.py", line 109, in _tailwind_class_manifest
    return " ".join(figure.dom_class_strings())
```

## Expected

The adapter should either:

- declare/enforce a compatible minimum `xy` version;
- feature-detect `dom_class_strings()` and treat a missing method as an empty optional class manifest; or
- provide a clear version-compatibility error before page compilation.

A static chart with no Tailwind class inventory should still compile and render.

## Context

This appeared while switching a Reflex dashboard from state-backed chart tokens to documented direct `Chart` sources. The page shell rendered, but live charts were blank due to a separate frontend/backend version mismatch. Static payload mode was the intended resilient fallback, but compilation then failed on this missing method.

## Environment

- `reflex-xy` 0.1.0 (editable repository package)
- installed `xy` core exposing `Figure` without `dom_class_strings()`
- Reflex 0.9.7
- macOS Apple Silicon

FarhanAliRaza:
### Motivation
- Prevent Reflex page compilation from crashing when an installed `xy` core `Figure` does not expose the optional `dom_class_strings()` API so static-chart fallback remains usable.

### Description
- Feature-detect `Figure.dom_class_strings` in `python/reflex-xy/reflex_xy/component.py` and return an empty Tailwind manifest when the method is absent.
- Add `tests/reflex_adapter/test_component.py::test_static_chart_without_class_inventory_still_compiles` to cover legacy `Figure`-like objects that lack `dom_class_strings()`.

### Testing
- Ran `PYTHONPATH=python/reflex-xy:python python -m pytest -q tests/reflex_adapter/test_component.py`, which passed (`8 passed`).
- Ran `python -m ruff check` and `python -m ruff format --check` on the modified files with no failures.
- Attempting `PYTHONPATH=python/reflex-xy:python python -m pytest -q tests/reflex_adapter` could not collect the full adapter suite due to missing optional test dependencies (`uvicorn` and `hypothesis`).

------

Closes #81

### DIFF
--- a/python/reflex-xy/reflex_xy/component.py
+++ b/python/reflex-xy/reflex_xy/component.py
@@ -105,8 +105,14 @@ def _tailwind_class_manifest(figure: Any) -> str:
     The inventory itself is core-Figure knowledge and lives on
     :meth:`xy.Figure.dom_class_strings`; reading the built figure avoids a
     second payload compilation (which can be expensive for large charts).
+    Older xy releases do not expose that optional inventory; their static
+    payloads remain renderable, but cannot contribute classes to Tailwind's
+    source scan.
     """
-    return " ".join(figure.dom_class_strings())
+    dom_class_strings = getattr(figure, "dom_class_strings", None)
+    if not callable(dom_class_strings):
+        return ""
+    return " ".join(dom_class_strings())
 
 
 def chart(source: Any, **props: Any) -> Any:
--- a/tests/reflex_adapter/test_component.py
+++ b/tests/reflex_adapter/test_component.py
@@ -110,6 +110,20 @@ def test_static_chart_classes_are_visible_to_tailwind_source_scan(app_cwd):
         assert class_string in rendered
 
 
+def test_static_chart_without_class_inventory_still_compiles(app_cwd):
+    """Older core Figures predate the optional Tailwind class inventory."""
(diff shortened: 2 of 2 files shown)


FarhanAliRaza (commented):
This is not needed as simple upgrade will solve this issue we might to add minimum version required so this does not happen again.

PR #103, opened 2026-07-20
Add pan/zoom interaction switches for static-viewport charts

### LINKED ISSUE #80: Add first-class controls to disable pan and zoom
## Problem

XY charts are interactive by default, but `interaction_config()` only configures hover, click, selection, brush, crosshair, view-change events, and linked views. There is no supported way to make a chart's viewport static.

As a result, dashboard charts that are intended as fixed summaries can still be dragged or wheel-zoomed, leaving bars or axes partially out of view. Hiding the modebar does not disable those gestures, and disabling pointer events on a framework wrapper is brittle because the chart's canvas owns the interaction handlers.

## Reproduction

```python
chart = xy.bar_chart(
    xy.bar(["A", "B", "C"], [10, 20, 30]),
    xy.modebar(show=False),
)
```

Plain-drag still pans and the wheel still zooms.

## Proposed API

```python
xy.interaction_config(pan=False, zoom=False)
```

- `pan=False`: ignore plain-drag pan gestures.
- `zoom=False`: ignore wheel, box-zoom, double-click reset, and zoom controls.
- Defaults remain enabled for backward compatibility.
- Hover and selection should remain independently configurable.

This would let dashboards use fixed-domain summary charts while keeping deliberately exploratory charts interactive.

FarhanAliRaza:
## Problem

XY charts are interactive by default, but `interaction_config()` only configured hover, click, selection, brush, crosshair, view-change events, and linked views — there was no supported way to make a chart's viewport static. Dashboard charts intended as fixed summaries could still be drag-panned or wheel-zoomed, leaving bars or axes partially out of view. Hiding the modebar does not disable the gestures, and disabling pointer events on a framework wrapper is brittle because the chart's canvas owns the interaction handlers.

## API

```python
chart = xy.bar_chart(
    xy.bar(["A", "B", "C"], [10, 20, 30]),
    xy.interaction_config(pan=False, zoom=False),
)
# or equivalently as chart-level kwargs: xy.bar_chart(..., pan=False, zoom=False)
```

- `pan=False` — plain-drag pan gestures are ignored; the modebar Pan button is hidden. Shift-drag / box select stay governed by `select`/`brush`.
- `zoom=False` — wheel zoom, box-zoom, and double-click reset are ignored; the modebar zoom controls (Zoom In/Out, Box Zoom, Reset View, percent label) are hidden. Double-click still clears selection.
- Defaults remain enabled: the flags serialize only when explicitly set, and the client treats an absent key as enabled, so existing specs are byte-identical.
- Hover, click, and selection remain independently configurable and functional on a static chart.

## Behavior details

- With `zoom=False` the wheel handler returns **before** `preventDefault()`, so the page scrolls naturally over a static chart embedded in a dashboard.
- With pan disabled, the drag-tracking object is never created; a stationary press still fires click/pick events and hover is unaffected.
- A chart with **both** flags disabled ignores inbound linked-view (`link_group`) updates; with only one disabled, linked updates still apply.
- Non-bool values raise `ValueError` via the existing centralized interaction validation.
- No Rust/ABI changes — interaction config never touches the core. `python/xy/static/` regenerated via `node js/build.mjs`.

## Testing

- New tests: explicit `pan`/`zoom` serialization (True and False), chart-level kwargs, and `ValueError` on non-bool values (`tests/test_components.py`).
- Existing exact-dict interaction spec tests pass unmodified aside from the deliberately extended declarative-contract test — confirming defaults serialize nothing.
- Full suite: 2021 passed. `ruff check`, `ruff format --check` clean; `ty check` has one pre-existing diagnostic unrelated to this change (present on a clean tree).

Closes #80


Alek99 (commented):
Review findings:

- [P2] `js/src/50_chartview.js:914`: charts with pan and zoom disabled still set `touch-action:none`, so a touchscreen swipe over a static chart cannot scroll the containing dashboard. Make touch-action depend on enabled gestures/drag mode.
- [P3] `js/src/53_interaction.js:1205`: pan-disabled charts still initialize `dragMode="pan"` and show a grab cursor even though plain dragging does nothing. Initialize/sanitize the mode from the enabled interactions.

PR #104, opened 2026-07-20
Fix FacetChart notebook display CSS isolation

### LINKED ISSUE #90: FacetChart notebook display leaks global CSS into the page (no widget path; _repr_html_ returns a standalone document)
## Summary

Displaying a `FacetChart` in Jupyter breaks the notebook's layout: the facet grid spills across cell boundaries and the surrounding cells look wrong. Regular `Chart`s display fine.

## Repro

```python
import numpy as np
import xy

rng = np.random.default_rng(42)
g = np.array(["A", "B", "C", "D"])[rng.integers(0, 4, 8000)]
x = rng.normal(0, 1, 8000)
y = x * np.array([0.2, 0.8, -0.5, 1.5])[np.searchsorted(["A", "B", "C", "D"], g)] + rng.normal(0, 0.4, 8000)

xy.facet_chart(
    xy.scatter(x="x", y="y", size=3.0, opacity=0.5),
    by="g", cols=2,
    data={"x": x, "y": y, "g": g},
    width=900, height=500,
)
```

Run in a notebook → the output renders but leaks outside its cell, visually corrupting neighboring cells.

## Root cause

1. **No widget display path.** `Chart` displays via `_ipython_display_` → the anywidget `FigureWidget`, contained in the output area. `FacetChart` only implements `_repr_html_` (`components.py:3293`), which delegates to `FacetGrid.to_html()` (`facets.py:170`) — a **full standalone HTML document** (`<!doctype html><html><head>…`). Jupyter injects that markup inline into the output div, so its document-level stylesheet applies to the *notebook page*:

   ```css
   html,body{margin:0;width:100%;min-height:100%;font-family:system-ui,sans-serif;background:#fff;}
   ```

   plus full-width grid styles — hence the spill.

2. **Compounding: `height` is per-panel for facets.** `FacetGrid.panel_height` returns `self.height` verbatim (`facets.py:154`) while `width` is total and divided across columns (`panel_width`, `facets.py:150`). `height=500` with 2 rows quietly composes a ~1012px grid. The width/height asymmetry is easy to hit and isn't documented in `facet_chart`'s docstring.

## Suggested fix

- Give `FacetChart` a real notebook path: `_ipython_display_` that either composes panel `FigureWidget`s in a container widget, or at minimum wraps the standalone doc in an `<iframe srcdoc=…>` so its CSS stays sandboxed. (The iframe wrapper is a working user-side workaround today.)
- Scope the standalone doc's CSS to a root class instead of `html,body` so inline injection degrades gracefully.
- Either make facet `height` mean total height (divided across rows, symmetric with `width`) or call out the per-panel semantics in the `facet_chart` docstring.

Environment: Jupyter (ipywidgets/anywidget transport), Linux, `xy` editable install of `main` @ 541a7da, Python 3.14.

FarhanAliRaza:
### Motivation

- FacetChart rendered a standalone HTML document via `_repr_html_` which injected document-level CSS into notebook pages and broke surrounding layout. 
- The facet `height` was documented ambiguously: it behaved as per-panel height while `width` was the total grid width, leading to unexpected composed heights. 
- A robust notebook display path should sandbox the facet document and provide an explicit IPython display path so frontends do not apply global styles.

### Description

- Change notebook representation to embed the standalone facet document in a sandboxed `srcdoc` iframe by returning `export.notebook_iframe(grid.to_html(), width=..., height=...)` from `FacetChart._repr_html_`. 
- Add an explicit `_ipython_display_` implementation that emits the isolated HTML payload using `display({...}, raw=True)` for rich display frontends. 
- Scope the facet standalone CSS beneath a root `.xy-facet-document` class and set the document `<body>` to that class so global selectors like `html,body{...}` no longer leak into host pages. 
- Clarify `facet_chart` docstring to state that `width` is the total grid width while `height` is the per-panel height, and add regression tests for the iframe wrapper, sandboxing, scoped CSS, and IPython display behavior in `tests/test_facets.py`.

### Testing

- Ran `PYTHONPATH=python pytest -q tests/test_facets.py`, which passed (`25 passed`).
- Ran `ruff check python/xy/components.py python/xy/facets.py tests/test_facets.py`, which passed with no issues. 
- Ran `git diff --check` to ensure no trailing/whitespace problems and verified working tree status.

------

Closes #90

### DIFF
--- a/python/xy/components.py
+++ b/python/xy/components.py
@@ -3384,7 +3384,18 @@ class FacetChart(Component):
         return self.to_html(path, custom_css=custom_css)
 
     def _repr_html_(self) -> str:
-        return self.to_html()
+        grid = self.figure()
+        return export.notebook_iframe(
+            grid.to_html(),
+            width=grid.width,
+            height=grid.grid_height + grid._title_height,
+        )
+
+    def _ipython_display_(self) -> None:
+        """Display the isolated facet document in notebook frontends."""
+        from IPython.display import display  # type: ignore[import-not-found]
+
+        display({"text/html": self._repr_html_()}, raw=True)
 
     def to_svg(self, path: Optional[str | PathLike[str]] = None) -> str:
         """A static SVG render of the facet grid (written to ``path`` if given)."""
@@ -4104,7 +4115,9 @@ def facet_chart(
         share_x: Whether panels share an x domain.
         share_y: Whether panels share a y domain.
         gap: Gap between panels in pixels.
-        **props: Additional shared chart properties.
+        **props: Additional shared chart properties. ``width`` is the total
+            grid width, while ``height`` is the height of each panel; the
+            composed height grows with the number of facet rows.
     """
     return FacetChart(
         children, by=by, cols=cols, share_x=share_x, share_y=share_y, gap=gap, **props
--- a/python/xy/facets.py
+++ b/python/xy/facets.py
@@ -204,12 +204,12 @@ class FacetGrid:
 <meta http-equiv="Content-Security-Policy" content="{export._STANDALONE_CSP}">
 <title>{title}</title>
 <style>
-html,body{{margin:0;width:100%;min-height:100%;font-family:system-ui,sans-serif;background:#fff;}}
-.xy-facet-title{{height:{self._TITLE_H}px;line-height:{self._TITLE_H}px;font:600 14px system-ui,sans-serif;margin:0;text-align:center;color:#1e293b;}}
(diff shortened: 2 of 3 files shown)


FarhanAliRaza:
https://github.com/reflex-dev/xy/pull/104 maybe need to merge in here to actually test because of css leaking the fix is hard to test but i have tested it locally after meging and it works.

Alek99 (commented):
Review finding:

- [P2] `python/xy/components.py:3388`: moving FacetChart notebook display into `notebook_iframe()` puts the toolbar in `sandbox="allow-scripts"` without `allow-downloads`. Browsers block the toolbar's PNG/SVG/CSV downloads in that sandbox. Add the narrowly scoped `allow-downloads` permission if notebook exports remain supported.

PR #105, opened 2026-07-20
facet_chart: add opt-in linked panel interactions

### LINKED ISSUE #92: facet_chart: opt-in linked zoom/pan/selection across panels (default stays independent)
## Summary

Feature request: facet panels should support **linked interactions** — synchronized zoom/pan (and optionally box-selection) across panels — behind an explicit opt-in. The default should remain what it is today: every panel zooms, pans, and selects independently.

## Current behavior

`facet_chart(share_x=True, share_y=True)` only aligns the panels' *initial* domains at build time (`components.py:3209` computes shared dims once). At runtime each panel is an independent `renderStandalone` instance (`facets.py:170`), so:

- zooming/panning one panel leaves the others at their original view — comparing the same region across panels means manually reproducing the same box zoom N times;
- box-select is per-panel; there's no way to brush one panel and see the same x-range (or the selected rows) highlighted in the rest.

There is no API to change this.

## Proposed API

```python
xy.facet_chart(
    xy.scatter(x="x", y="y"),
    by="g",
    cols=2,
    data=df,
    link="x",        # None (default) | "x" | "y" | "both"
    link_select=True # default False: brushing one panel echoes the selection to all
)
```

- `link=None` (default): today's behavior, fully independent panels.
- `link="x" | "y" | "both"`: a zoom/pan in any panel broadcasts its new domain to the other panels on the linked dimension(s). Natural pairing with `share_x`/`share_y` — linking a dimension probably implies sharing its initial domain.
- `link_select` (or fold into `link`): a brush in one panel applies the same data-space predicate across panels, and `on_select` fires once with the union.

## Notes

- Double-click reset should reset all linked panels together.
- The render client already recomputes per-view domains on zoom; linking is a broadcast of the view window between panel instances in the same grid (all panels live in one document/widget, so no kernel round-trip is required for the view sync itself).
- Related: #90 (facet display path) — if facets grow a proper widget path, the link channel could ride the same container.

Environment/context: `xy` main @ 541a7da, playground testing in Jupyter.

FarhanAliRaza:
### Motivation

- Facet panels should be able to synchronize navigation and brushing so users can compare the same data window or selection across panels while preserving the current default of independent panel interactions.

### Description

- Add `link: Optional[str] = None` and `link_select: bool = False` to `facet_chart`/`FacetChart` with eager validation of allowed values (`None`, `"x"`, `"y"`, "both").
- Make linked axes share their initial domain at build time and set a browser-local `link_group` + `link_axes` on per-panel `Figure` instances when linking is requested, preserving existing `share_x`/`share_y` behavior for initial-domain alignment.
- Extend the client to broadcast and receive linked selection predicates (range, polygon, and clear) over the existing `BroadcastChannel`, add `_broadcastLinkedSelection`, and update selection handlers to accept `opts` that suppress duplicate dispatch/broadcast when applying incoming linked selections.
- Surface the new options in docs (`docs/guides/dashboards-and-linked-views.md`) and add/adjust tests covering independent defaults, explicit link choices, link-implied domain sharing, invalid `link` values, and selection linking logic.

### Testing

- Built client bundle with `node js/build.mjs` and ran targeted suites `PYTHONPATH=python pytest -q tests/test_facets.py tests/test_components.py tests/test_static_client_security.py`, which all passed (138 tests).
- Ran the full test suite with `PYTHONPATH=python pytest -q`, which succeeded (1,835 passed, 77 skipped).
- Verified static client contract tests and fixed a transient client-regression discovered during the first run before the final verification, after which the security/contract tests passed.

------

Closes #92


Alek99 (commented):
Review findings:

- [P1] `python/xy/components.py:3306`: `link=None` removes the runtime linking that existing `share_x=True/share_y=True` facets receive today. This is a breaking change against the prior tests/docs; preserve it or provide an explicit migration/versioning path.
- [P1] `js/src/50_chartview.js:442`: incoming linked selections use `_selectLocal*`, which skips density traces and cannot select drilled GPU data. Receiver facets do not highlight, and selection callbacks only see the source facet rather than the requested cross-panel result.
- [P2] `js/src/50_chartview.js:458`: selection-only links still broadcast/apply view events with an empty `link_axes`, producing redundant LOD requests and callbacks on untouched facets. Avoid broadcasting view updates when no axes are linked.

PR #108, opened 2026-07-20
Add hard viewport bounds that clamp pan and zoom

### LINKED ISSUE #82: Add hard viewport bounds that clamp pan and zoom
## Problem

An explicit axis `domain=(lo, hi)` defines the initial visible range, but it does not constrain later interaction. Users can pan the plot completely or partially outside that domain, leaving a chart half-visible or empty.

Disabling pan and zoom entirely addresses static charts (#80), but exploratory charts need a middle ground: allow navigation while preventing the camera from leaving meaningful data bounds.

## Reproduction

```python
chart = xy.bar_chart(
    xy.bar(["A", "B", "C"], [10, 20, 30]),
    xy.x_axis(domain=(0, 30)),
)
```

Plain-drag can move the viewport beyond `(0, 30)`; the domain is treated as an initial view rather than a hard boundary.

## Proposed API

Support separate navigation bounds, for example:

```python
xy.x_axis(domain=(0, 30), bounds=(0, 30))
xy.y_axis(bounds="data")
```

Expected behavior:

- Pan is clamped so the visible window cannot leave the configured bounds.
- Zoom-out is clamped to the maximum bounded extent.
- Zoom-in remains available within the bounds.
- Reversed and log axes preserve their existing semantics.
- `domain` remains the initial view for backward compatibility.
- `bounds=None` preserves today's unrestricted navigation.

This is distinct from disabling interaction: it keeps exploratory charts useful without allowing users to lose the data.

FarhanAliRaza:
### Motivation
- Provide a way to keep interactive navigation (pan/zoom) inside meaningful data extents separate from the initial `domain` so users cannot pan/zoom the camera completely away from the chart.
- Keep exploratory interaction available while preventing empty/partially-visible charts and preserve existing semantics for reversed and log axes.

### Description
- Add a `bounds` field to `Axis` and expose `bounds=(lo, hi)` or `bounds="data"` on `xy.x_axis()` and `xy.y_axis()` so navigation limits are declaratively configurable.
- Validate `bounds` (including log-axis positivity) on the Python side and serialize `spec["bounds"]` separately from `domain`, resolving `"data"` to the data range at payload build time.
- Clamp the initial viewport, drag panning, wheel/box/button zooms, animated view transitions, and linked-view updates on the client by adding `_clampAxisRange` and `_clampView` helpers and integrating them where views are set/updated; clamping works in scale coordinates to preserve reversed and log semantics.
- Add `tests/test_viewport_bounds.py` covering payload serialization, `"data"` resolution, validation, and presence of client clamping hooks; regenerate the committed browser bundles so the shipped static client contains the clamp logic.

### Testing
- Ran unit tests: `PYTHONPATH=python python -m pytest tests/test_viewport_bounds.py tests/test_type_surface.py tests/test_figure.py -q` and `PYTHONPATH=python python -m pytest tests/test_viewport_bounds.py -q`, both succeeded (final run: 166 passed, 1 skipped; viewport-bounds tests passed).
- Ran formatting and lint checks: `ruff format --check` / `ruff check` on the modified files and fixed import ordering; all checks passed.
- Rebuilt the static client: `node js/build.mjs --check` and verified the generated bundles; check passed.
- Ran `git diff --check` to ensure no whitespace/check issues; no problems found.

------

Closes #82

### DIFF
--- a/examples/demo-advance.ipynb
+++ b/examples/demo-advance.ipynb
@@ -270,6 +270,33 @@
     ")\n",
     "chart"
    ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "viewport-bounds-demo",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# PR verification: the initial viewport is smaller than the hard navigation bounds.\n",
+    "# Drag and wheel-zoom aggressively; neither axis should move beyond the red rules.\n",
+    "viewport_x = np.linspace(0, 100, 501)\n",
+    "viewport_y = np.sin(viewport_x / 7) + 0.15 * np.cos(viewport_x / 2)\n",
+    "\n",
+    "xy.line_chart(\n",
+    "    xy.line(viewport_x, viewport_y, name=\"signal\", color=\"#5477c4\", width=3),\n",
+    "    xy.vline(0, text=\"x min\", color=\"#ef4444\", width=2),\n",
+    "    xy.vline(100, text=\"x max\", color=\"#ef4444\", width=2),\n",
+    "    xy.hline(-1.25, text=\"y min\", color=\"#ef4444\", width=2),\n",
+    "    xy.hline(1.25, text=\"y max\", color=\"#ef4444\", width=2),\n",
+    "    xy.x_axis(label=\"x (hard bounds: 0 to 100)\", domain=(30, 60), bounds=(0, 100)),\n",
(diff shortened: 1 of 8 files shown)


Alek99 (commented):
Review findings:

- [P1] `js/src/50_chartview.js:216`: the initial viewport is clamped only in the browser, after density/decimation payloads and static-export ranges have been built. For `domain=(0, 1_000_000), bounds=(0, 1)`, a 0..1 scatter is still binned/exported over 0..1_000_000. Clamp before payload generation, or reject incompatible domain/bounds.
- [P2] `python/xy/_figure.py:1003`: bounds changes renderer semantics but ships under protocol 3, so older clients silently ignore it and allow unrestricted navigation. Bump the protocol and regenerate bundles.

PR #109, opened 2026-07-20
Add native zero-inclusive symlog axis scale

### LINKED ISSUE #83: Add a native zero-inclusive long-tail axis scale
## Problem

XY supports linear and logarithmic numeric axes, but common long-tailed measures often contain valid zeros. A linear axis compresses nearly all observations near the origin, while a log axis cannot represent zero.

In a 2.19M-row payroll scatter, preserving zero-pay records while making the long tail legible required transforming every value with `log10(max(value, 0) + 1)`, supplying transformed tick positions manually, and inverting selection bounds back to dollars in application code.

That workaround leaks display transforms into data, hover, selection, linking, formatting, and domain configuration.

## Proposed API

Add a native zero-inclusive long-tail scale, such as `symlog` and/or configurable `log1p`:

```python
xy.x_axis(type_="symlog", constant=1_000, format="$,.0f")
xy.y_axis(type_="log1p", format="$,.0f")
```

Expected behavior:

- Zero remains representable.
- Positive long tails spread naturally.
- Optional negative values work with `symlog`.
- Ticks and hover expose original values, not transformed coordinates.
- Pan, zoom, selection, callbacks, and linked views all report original data-space bounds.
- Direct and density scatter tiers use identical scale semantics.
- Static SVG/PNG and browser rendering remain consistent.

A native scale would remove error-prone application transforms and keep exact interaction semantics intact.

FarhanAliRaza:
### Motivation
- Long-tailed numeric measures commonly include valid zeros which a `log` axis cannot represent while a `linear` axis compresses the tail, so a zero-inclusive long-tail scale (symmetric log / `symlog`) is needed.
- The goal is to provide a native axis type that preserves original data-space values for ticks, hover, selection, pan/zoom, density rendering, and payloads without leaking display transforms into application logic.

### Description
- Add `symlog` as an accepted axis `type_` and a `constant` parameter to `x_axis()`/`y_axis()` and `Figure.set_axis()` to configure the linear region width around zero, including validation of a positive constant. (`python/xy/components.py`, `python/xy/_figure.py`)
- Serialize `scale: "symlog"` and `constant` into axis payloads and ensure `Figure._axis_spec` emits `scale`/`constant` while preserving existing semantics for `linear`/`log`. (`python/xy/_figure.py`)
- Implement symmetric-log transforms in the static PNG/SVG pipeline via `_Scale` and tick generation that produce ticks in original data units (including zero). (`python/xy/_svg.py`)
- Wire `symlog` through the browser-side code: add `symlog` branch to `axis` mapping, implement JS coordinate transform and inverse, propagate a per-axis `constant` uniform into GLSL, and update shader map/calls so WebGL rendering, density, hit/pick, and interaction layers operate using the symlog transform consistently. (`python/xy/static/index.js`, `python/xy/static/standalone.js`)
- Add documentation for the zero-inclusive long-tail scale and an integration test exercising API, payloads, static transforms, and ticks. (`docs/core-concepts/axes-and-scales.md`, `tests/test_symlog_axis.py`)

### Testing
- Ran focused and integrated Python tests: `PYTHONPATH=python python -m pytest tests/test_symlog_axis.py -q` (8 passed) and `PYTHONPATH=python python -m pytest tests/test_figure.py tests/test_svg_export.py tests/test_png_export.py -q` (212 passed, 1 skipped), and `PYTHONPATH=python python -m pytest tests/test_components.py -q` (76 passed).
- Static JavaScript validation succeeded with `node --check python/xy/static/index.js` and `node --check python/xy/static/standalone.js` and `git diff --check` reported no issues.
- Browser conformance / screenshot smoke could not be run here because Playwright’s Chromium executable is not installed in this environment (so end-to-end screenshot verification was not executed).

------

Closes #83


FarhanAliRaza:
image
without 
vs 

with 
image

Alek99 (commented):
Review findings:

- [P1] `js/src/40_gl.js:76-78`: symlog is applied after decoding offset-encoded f32 coordinates. With `x=[0, 1, 1e12]` and `constant=1`, 0 and 1 collapse to the same decoded value even though symlog should visibly separate them. Encode geometry in scale coordinates or retain sufficient precision before the transform.
- [P1] `js/src/40_gl.js:348`: density aggregation remains uniform in raw data space. Clusters at x=1 and x=1000 can land in the same raw grid cell even though they are far apart in symlog display space. Bin/decimate in scale coordinates.
- [P2] `python/xy/_svg.py:2260`: static heatmap/density grids are stretched only between transformed endpoints, so internal cells are not symlog-transformed. PNG follows the same pattern.
- [P2] `python/xy/_figure.py:993`: `scale="symlog"` changes renderer semantics but the protocol remains 3; an older client will silently treat it as linear. Bump the protocol and regenerate bundles.

PR #112, opened 2026-07-20
ENG-10442: Update style docs

carlosabadia:


Alek99:
I found three actionable issues in this PR:

1. **Coverage tests validate unpublished documentation.** `DOCS_CONFIG` excludes `styling/chrome-slots.md`, `component-variations.md`, and `mark-styles.md`, but the coverage tests still read those files directly. CI therefore stays green even though the published docs no longer document the `root`, `chrome`, `canvas`, `labels`, `badge`, or `badge_item` slots. The reference material should move into the new canonical pages, and the tests should inspect discovered/published pages.

2. **The invisible tooltip trace breaks the accessible series summary.** The transparent scatter helper in `docs/styling/examples.md` is still counted as a data series. In the live page the accessible text says “3 data series. Series: Desktop, Mobile,” although only two series are presented. Please bind the tooltip data without another trace, or exclude helper traces from the accessibility summary.

3. **Permanent WebGL rebuild failures retry forever.** The catch in `_rebuildEvictedContext()` treats every initialization, shader, payload, or draw failure as temporary context pressure and schedules another rebuild indefinitely. A permanent failure will replace the canvas and create another WebGL context every second without surfacing `context_restore_failed`. Please reuse the transient-error classification from the restored-context handler, or cap retries and report the failure.

I verified the current head `383d930`. Locally, the docs suite (71 tests), production validators, Ruff, static bundle freshness, and benchmark-environment tests all pass.

PR #113, opened 2026-07-20
ENG-10441: Complete Reflex events and cross-filtering

FarhanAliRaza:
Closes the P0 interaction loop for `reflex_xy`: browser interaction → semantic event → Reflex handler → State update → dependent `@reflex_xy.figure` recompute → republish behind stable tokens, with no remount and no feedback loops.

## What changed

**Event envelopes (v1).** `on_point_hover`, `on_point_click`, `on_select_end`, and `on_view_change` now deliver versioned, JSON-bounded envelopes carrying the stable figure `token`, trace identity, the canonical row id (never a shipped/decimated buffer index), `data`/`datum`, and — for clicks — canvas-relative screen coords and keyboard modifiers.

**Bounded selections + complete data (ENG-10461).** Selection replies opt into `include_rows` (sent only when `on_select_end` is wired) and carry `kind`/`mode`/`bounds`/`polygon`/`canonical_row_ids`/`rows`/`truncated`, capped at `SELECTION_EVENT_ROW_LIMIT = 1000` rows / `SELECTION_EVENT_ID_LIMIT = 10000` ids with `total_count` always exact. New `Selection.rows()` gives a deterministic, JSON-safe canonical projection (pick-result shape); new `reflex_xy.resolve_selection(event)` re-resolves the geometry server-side against canonical f64 columns for the complete, unbounded selection — large selections never ride the browser as JSON.

**Event enablement, throttling, loop hygiene.** The JSX wrapper injects `click`/`view_change` interaction flags only when those handler props are wired (zero overhead otherwise), throttles hover dispatch (120 ms latest-wins), debounces view changes (200 ms trailing) and suppresses non-user sources. On a stable-token republish the wrapper preserves the user's viewport and re-issues the active selection without redispatching handlers, so cross-filter refreshes can't loop or reset zoom.

**Stale-event rejection.** Clients echo the last payload version (`v`) on every kernel message; `XYNamespace.on_msg` drops messages minted against a replaced figure version (same pattern as `drill_seq` staleness). Clients that omit `v` remain compatible.

**Input parity.** Keyboard Enter/Space on a focused datum emits the same click contract (screen + modifiers) as pointer activation; render-client bundles regenerated via `node js/build.mjs`.

**Docs & demo.** `docs/engineering/design/reflex-integration.md` §5.1 documents the envelopes, limits, truncation fallback, static-`src` caveats, and shared-handler/viewport-sync patterns; the demo app cross-filters by selection rows and demonstrates `resolve_selection`.

## Verification

- `uv run pytest` — 1949 passed, 65 skipped (includes new unit + socket-E2E coverage: enriched selection round-trip, stale-version drop, `resolve_selection`, keyboard-activation contract, payload-bound tests)
- `uv run ruff check .` / `ruff format --check .` / `uv run ty check` — clean
- `node js/build.mjs` — bundles fresh
- `uv run --with pre-commit pre-commit run --all-files` — all hooks pass

Not covered: the browser smoke (`scripts/reflex_ws_smoke.py`) needs a running demo app + headless Chromium; worth a manual pass before release.


tests/reflex_adapter/test_selections.py
```diff
@@ -0,0 +1,53 @@
...
+            token,
+            {
+                "kind": "lasso",
+                "polygon": [[-1, -1], [4, -1], [2, 4]],
+                "total_count": 3,
+            },
+        )
+    )
+    assert lasso is not None
+    assert len(lasso) > 0
```
masenf: the chart data is deterministic, we should have a better assertion here

tests/reflex_adapter/test_socket_data_plane.py
```diff
@@ -193,6 +193,76 @@ async def main():
...
+            )
+            reply = await collector.next(collector.messages)
+            await client.disconnect()
+        message = reply["message"]
+        assert message["version"] == 1
+        assert message["kind"] == "box"
+        assert message["rows"][0]["index"] == 0
+        assert message["canonical_row_ids"][0]["ids"] == list(range(8))
+
+    run(main())
```
masenf, line 229: is there a reason these aren't just async test functions?

masenf (commented):
it would be nice to have TypedDict declarations for the `event` payloads handled on the reflex side

tests/reflex_adapter/test_socket_data_plane.py
FarhanAliRaza: we dont have pytest-asyncio installed. Maybe add it inanother pr.

masenf (approved):

PR #115, opened 2026-07-20
Add export format parity and a unified export API (ENG-10447)

Alek99:
Closes [ENG-10447](https://linear.app/reflex-dev/issue/ENG-10447/add-export-format-parity-and-a-unified-export-api-to-xy).

Brings XY's export surface to practical format parity with Plotly — PNG, JPEG/JPG, WebP, SVG, PDF, and interactive HTML — behind one unified API, while preserving the browser-free native fast path for **every** format.

## API

```python
chart.to_image("pdf", width=1200, height=800, scale=2)     # bytes
chart.write_image("reports/revenue.webp")                  # inferred + atomic
xy.write_images(figures=[overview, detail],
                files=["overview.svg", "detail.pdf"])      # mixed batch

xy.chart(
    xy.line("date", "revenue", data=frame),
    xy.export_config(formats=["png", "webp", "svg", "csv"],
                     filename="revenue", width=1200, height=800, scale=2),
)
```

`to_png` / `to_svg` / `to_html` are unchanged compatibility conveniences.

## Format & engine matrix

| Format | Native (browser-free) | Chromium | Notes |
| --- | --- | --- | --- |
| png | ✅ default | ✅ | transparency |
| jpeg/jpg | ✅ default — new pure-numpy baseline encoder (`_jpeg.py`, ~63 ms @ 1800×1000) | ✅ CDP screenshot | no alpha → flattens onto background (default white), `quality` 1–100 |
| webp | ✅ default — new bit-exact **lossless** VP8L encoder with alpha (`_webp.py`) | ✅ CDP screenshot (lossy, honors `quality`) | documented lossless-native policy |
| svg | ✅ always | — | native-only (a screenshot can't be vector) |
| pdf | ✅ default — new vector backend (`_pdf.py`) converting XY's own SVG subset | ✅ `printToPDF` | vector text (Helvetica AFM metrics), axial-shading gradients incl. alpha soft-masks, density/heatmap layers embedded as bounded rasters (hybrid-vector policy); strict whitelist so SVG-generator drift fails loudly |
| html | ✅ `to_html` | — | `write_image("chart.html")` routes there |

`engine="auto"` is deterministic: native for everything, Chromium only when `custom_css` needs a real CSS engine. Missing-browser errors name `XY_BROWSER` and the supported executables. Batch export shares **one** Chromium session across all browser-resolved files.

## Background & quality policy

`background="auto" | <css color> | "transparent"` behaves identically across raster/vector/browser output; JPEG rejects `transparent` instead of silently flattening. An explicit color now paints the same single backdrop in the rasterizer and SVG (plot-rect default harmonized). `quality` applies to JPEG and Chromium-WebP; native WebP is always lossless by policy.

## Modebar / declarative / Reflex

`export_config` rides the spec: the modebar download menu is now config-driven (client-safe subset png/jpeg/webp/svg/csv, ordered; `formats=[]` hides it) with the same filename/scale/background/quality semantics as Python — verified working in kernel-free standalone HTML via headless CDP (menu items, configured filename, correct extensions, toBlob-fallback naming). Reflex charts inherit the same spec-driven config.

## Facets

`FacetGrid`/`FacetChart` gain `to_image`/`write_image` over the same matrix: native raster composes panel renders; SVG/PDF compose vector panels; Chromium renders the HTML grid.

## Verification

- 2018 passed / 5 skipped full suite; new suites: `test_jpeg.py` (23), `test_webp.py` (19, bit-exact PIL round-trips incl. alpha), `test_pdf_export.py` (xref-offset parser, vector-text assertions, image XObjects, determinism), `test_image_export.py` (31, full matrix + engine/background/quality/batch/config/facet + fake-CDP chromium plumbing).
- Pillow added to **dev** extras only, as the encoder decode oracle.
- Native PDF cross-validated against two independent renderers (macOS Quartz + headless Chromium pixel-diffs of 20 figure kinds).
- `render_smoke_nonumpy.py`, `abi_smoke.py`, pre-commit hooks, ruff check/format, `ty` (no new diagnostics), `cargo test` all green. No Rust changes; JS bundle rebuilt via `node js/build.mjs`.

Follow-ups deliberately out of scope: EPS (dropped by modern Plotly/Kaleido too), golden-image CI corpus for the new formats, benchmark harness rows for JPEG/WebP/PDF export.


pyproject.toml
```diff
@@ -29,6 +29,9 @@ dev = [
     # Optional *input* format, never a runtime dependency: the Arrow ingest
     # tests importorskip it, so it lives in dev where CI exercises them.
     "pyarrow>=15",
+    # Decode oracle for the native JPEG/WebP encoders, never a runtime
+    # dependency: the image-export tests importorskip it.
+    "pillow>=10",
```
Alek99, line 34: [P1] Regenerate `uv.lock` after adding Pillow. The committed lockfile does not contain Pillow or include it in the `xy` dev extra, so `uv sync --frozen --extra dev` omits the decoder oracle and the new JPEG/WebP suites silently skip via `pytest.importorskip`. Please run `uv lock` and commit the result.

python/xy/_svg.py
```diff
@@ -1749,6 +1749,12 @@ def tick_span(style: dict[str, Any]) -> tuple[float, float, float]:
     # the --chart-bg token over the plot rect only. Solid colors only —
     # gradients stay browser-only, and an unset token stays transparent.
     backgrounds = ""
+    # Export-time canvas override (unified export API `background=`): one
+    # backdrop rect behind the figure patch. "transparent"/"none" mean "no
+    # backdrop", which is already SVG's default — nothing to paint.
+    canvas_paint = spec.get("canvas_background")
```
Alek99, line 1778: [P1] The explicit export backdrop is painted underneath the figure and plot theme backgrounds, so `background=` is invisible whenever `theme(background=...)` or `plot_background` is set. I reproduced `background="#112233"` producing the theme's red margins and green plot in both native and Chromium PNGs; `transparent` remains opaque too. Explicit export backgrounds need to replace/suppress those theme paints across the native SVG/PDF/raster and browser paths.

python/xy/export.py
```diff
@@ -506,61 +531,119 @@ def html_to_png(
...
+        if figs is not None:
+            raise ValueError("pass figs positionally or figures=, not both")
+        figs = figures
+    if files is not None:
+        if paths is not None:
+            raise ValueError("pass paths positionally or files=, not both")
+        paths = files
+    if figs is None or paths is None:
+        raise ValueError("write_images needs both figures and files")
+    figs = [f.figure() if callable(getattr(f, "figure", None)) else f for f in figs]
```
Alek99: [P2] Converting chart objects to `Figure` here drops `Chart._export_defaults`. A chart configured for 123×77 at scale 1 exports at 123×77 directly, but `write_images([chart], ...)` produces 600×400 from the chart's 300×200 dimensions and the batch scale default. Please resolve per-chart declarative defaults before discarding the wrapper.

python/xy/facets.py
```diff
@@ -325,6 +346,115 @@ def to_png(
...
+                )
+            canvas = self._compose_rgba(scale, background)
+            if fmt == "jpeg":
+                from . import _jpeg
+
+                return _jpeg.encode(export._flatten_alpha(canvas), quality=quality or 90)
+            from . import _webp
+
+            return _webp.encode(canvas)
+        doc = self.to_html(custom_css=custom_css)
```
Alek99: [P2] The validated facet `background` is never added to the HTML passed to Chromium; it is used only for the transparent CDP flag below. A real Chrome export with `background="#112233"` still had a white corner. Please build the document with the background override, as the single-chart path does, including PDF.

python/xy/components.py
```diff
@@ -2815,6 +2936,122 @@ def to_png(
...
+        quality: Optional[int],
+    ) -> dict[str, Any]:
+        """Fill omitted export options from the chart's `export_config`.
+
+        Direct arguments always win. Declarative defaults degrade gracefully
+        where a format cannot honor them (config quality is dropped for
+        non-lossy formats; a config "transparent" background is dropped for
+        JPEG) — only *explicit* arguments produce hard errors downstream."""
+        config = self.figure().export_options or {}
+        if quality is None and fmt == "jpeg":
```
Alek99: [P2] Declarative quality is copied only for JPEG. `export_config(quality=37)` followed by `to_image("webp", engine=Engine.chromium)` passes the generic default 90 to Chromium. Chromium WebP should inherit the configured quality while native lossless WebP continues ignoring it.

js/src/53_interaction.js
```diff
@@ -1058,7 +1074,9 @@ Object.assign(ChartView.prototype, {
       selectMenu.style.visibility = "visible";
     };
     setExportMenuOpen = (open, restoreFocus = false) => {
-      const show = Boolean(open);
+      // export_config(formats=[]) leaves nothing to show: the grip stays a
+      // pure drag handle rather than opening an empty menu.
+      const show = Boolean(open) && exportMenuItems.length > 0;
```
Alek99, line 1079: [P2] This makes `exportMenuItems` empty for `export_config(formats=[])` (and PDF/HTML-only lists), but the existing ArrowDown/ArrowUp handler still indexes that array and calls `.focus()` on `undefined`, causing a runtime `TypeError`. Please guard the keyboard handler when there are no client-side export items.

python/xy/export.py
```diff
@@ -506,61 +531,119 @@ def html_to_png(
...
+
+    # Resolve the whole plan before any I/O so bad arguments fail the batch
+    # up front instead of after a partial export.
+    plan: list[tuple["Figure", str | PathLike[str], str, str, Optional[int], Optional[str]]] = []
+    for fig, path, fmt in zip(figs, paths, fmts, strict=True):
+        if fmt == "html":
+            plan.append((fig, path, fmt, "html", None, None))
+            continue
+        resolved = _resolve_image_engine(engine, fmt, custom_css)
+        file_quality = (
```
Alek99, line 628: [P2] The batch contract says quality is ignored for non-lossy members, but a mixed native JPEG+WebP batch with `quality=50` fails here because native WebP is routed through `_validated_quality`. Apply quality to JPEG and Chromium WebP only; native lossless WebP should not abort an otherwise valid mixed batch.

Alek99 (commented):
I found seven actionable contract gaps in the new export paths. Existing CI and the full local suite pass, but the cases called out inline are not currently covered.

pyproject.toml
Alek99: Fixed in 49db00b — ran `uv lock`; Pillow now resolves in the lock (`pillow v12.3.0`) and is wired into the `xy` dev extra, so `uv sync --frozen --extra dev` installs the oracle and the JPEG/WebP suites run for real.

python/xy/_svg.py
Alek99: Fixed in 49db00b. Explicit `background=` now replaces the theme paints instead of layering under them: a shared `_svg.apply_export_background` (used by raster, SVG, and thereby PDF) drops the theme figure patch and turns `--chart-bg` transparent so the override composites exactly once; the browser paths inject the same override as `!important` CSS that beats the root's inline theme background and the `--chart-bg` token the client reads. Verified with your repro (theme red margins + green plot): native PNG/WebP/SVG and real-Chromium PNG all show the override color edge-to-edge, and `transparent` yields alpha-0 in both regions. Covered by `test_explicit_background_replaces_theme_paints` and `test_browser_background_css_overrides_theme_tokens`.

python/xy/export.py
Alek99: Fixed in 49db00b — `write_images` keeps the chart wrapper long enough to resolve its `_export_defaults` per file (width/height/scale/background/quality, including the engine-aware quality rule) before compiling to a `Figure`; batch-level arguments still override. Your 123×77@1 case is pinned in `test_write_images_resolves_chart_export_config_defaults`.

python/xy/facets.py
Alek99: Fixed in 49db00b — the facet browser branch now injects the validated background into the captured document via the same `_background_css` helper as the single-chart path (PDF included), and the native facet SVG passes the override into each panel so panel theme paints can't bury the grid backdrop. Verified against real Chromium (corner pixel = override color); pinned by `test_facet_browser_background_reaches_document`.

python/xy/components.py
Alek99: Fixed in 49db00b — `_export_defaults` now takes a `lossy_webp` flag derived from the resolved engine, so `export_config(quality=37)` reaches Chromium WebP while native WebP stays lossless and ignores it. Pinned by `test_export_config_quality_reaches_chromium_webp`.

js/src/53_interaction.js
Alek99: Fixed in 49db00b — the grip's ArrowDown/ArrowUp handler returns early when there are no client-side export items, matching the pointer path's guard. Verified via CDP with `export_config(formats=[])`: no page errors, menu stays closed.

python/xy/export.py
Alek99: Fixed in 49db00b — batch quality now applies to JPEG and Chromium-WebP members only; native lossless WebP no longer aborts a mixed batch (the value range itself is still validated once up front, so `quality=500` fails fast). Pinned by the extended `test_write_images_quality_ignored_by_non_lossy_batch_members`.

PR #118, opened 2026-07-21
Fix standalone density flicker on pan/zoom-out

adhami3310:
## Problem

In a kernel-less `to_html` density export, the density visibly flickered on the slightest drag — swapping between the full-data overview grid and the retained sample. Two distinct causes:

1. **Color/brightness flip on pan.** `_requestSampleRebin` re-binned the sample whenever the view wasn't *exactly* at the home extent, so a pure pan swapped the full-data overview grid for a sample-derived grid normalized against a much smaller maximum. The two normalize differently, so the density jumped.

2. **Points vanishing on pan/zoom-out.** The retained sample overlay was gated by `_viewInside(s.win)`, so its individual points dropped out the instant the view left the sample's home window, while the density stayed — an inconsistent "density loses its grain" effect.

## Fix

1. Gate the re-bin on view **span**, not position: pans and zoom-outs keep (and restore) the full-data overview; only a genuine **zoom-in** re-bins, where the sample actually adds resolution the overview lacks.

2. Add `_viewOverlaps` and draw the sample whenever it **overlaps** the view. Points are positioned in data space and the GPU clips off-screen ones, so the "density + points" look now travels with the view and only drops when panned entirely off the data.

Scoped strictly to the standalone (no-kernel) path; the kernel `density_view` path is unchanged.

## Tests

Adds `tests/test_density_pan_no_rebin.py` — a headless-Chromium regression test asserting:

- a pan keeps the overview and spawns **no** re-bin worker,
- a zoom-in still routes to the sample re-bin path,
- the sample overlay persists through a zoom-out but drops when panned entirely off the data.

Render smoke + related suites pass; bundles regenerated via `node js/build.mjs`; ruff/format clean.

### DIFF
--- a/js/src/50_chartview.js
+++ b/js/src/50_chartview.js
@@ -1825,7 +1825,12 @@ class ChartView {
 
   _drawDensitySample(g, x0, x1, y0, y1, opacityScale = 1) {
     const s = g && g.sampleOverlay;
-    if (!s || !s.n || !this._viewInside(s.win)) return;
+    // Draw the retained sample whenever it overlaps the view, not only when the
+    // view sits fully inside the sample window: a pan or zoom-out must keep the
+    // same "density + points" look instead of dropping the points the instant
+    // the view crosses the sample's home extent (the density grid stays, so the
+    // points have to as well). Off-screen points are clipped by the GPU.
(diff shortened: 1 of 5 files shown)


masenf:
@greptile review 🙏

PR #129, opened 2026-07-21
Unified view-state layer: spec + implementation (history, programmatic control, axis gestures, hover payload)

### LINKED ISSUE #110: View history stack so user can go back to previous state.

### LINKED ISSUE #111: Support custom tooltip renderers with payload and cursor coordinates in reflex_xy
## Summary

`xy.tooltip(render=...)` preserves the supplied component in core XY through
`chart.chrome_components()`, but the shipped `reflex_xy.chart(...)` adapter
does not mount or update it.

This makes it impossible to build a cursor-positioned custom tooltip similar
to Recharts' `<Tooltip content={...} />`.

## Current behavior

- `xy.tooltip(render=custom_tooltip)` accepts and retains the object.
- `reflex_xy.chart(...)` does not read `chart.chrome_components()`.
- `XYChart.jsx` has no custom tooltip renderer or tooltip-content prop.
- Live `on_point_hover` events receive only the resolved row dictionary.
- The event does not include cursor coordinates, chart-relative coordinates,
  active state, label, series metadata, or a structured payload.
- A host-rendered workaround can display the hovered row elsewhere, but it
  cannot reliably follow the pointer or behave like the built-in tooltip.
- Static charts cannot use the backend-event workaround.

The built-in XY tooltip continues to work; this issue concerns fully custom
framework-owned tooltip content.

## Expected behavior

Provide an adapter-supported custom tooltip contract comparable to Recharts:

```python
xy.tooltip(render=custom_tooltip)

### LINKED ISSUE #120: Support per axis scaling
when hovering over an axis, we could show the resize cursor (vertical for y, horizontal for x), and then click+drag or scroll would only change the range of the hovered axis

### LINKED ISSUE #121: handle out of band programmatic zoom/pan/select via reflex_xy
you don't want to have to recreate the chart to set the window, zoom level, or selected points; it should be possible to do this via the reflex_xy API, similar to .append

### PR DESCRIPTION
**Pull request by Alek99**
## Summary

Adds `spec/design/view-state.md` — the design for one canonical, serializable view-state object behind a single clamped mutation path — **and implements all four of its rollout phases in the same PR**. The spec's status is flipped to *implemented*; `spec/api/interaction.md` and `spec/design/wire-protocol.md` are updated as the shipped authorities, and view-state.md §12 records the implementation divergences.

Four open requests land as derivations of one primitive:

- #121 — **programmatic control**: `chart.set_view / reset_view / select / clear_selection / view_state` (notebook widget), and `reflex_xy.set_view(token, …)` et al. mirroring the `.append` path — figure lock → one wire message → room-wide apply, no re-render, no payload reship.
- #110 — **view history**: a client-local 64-entry ring of durable-state snapshots coalesced per gesture by `interaction_id`; modebar Back/Forward buttons; reset is navigation, so Back undoes a double-click. `interaction_config(history=…)` switches it.
- #120 — **per-axis gestures**: hover an axis band (tick strip + plot-side gutter) → resize cursor; wheel zooms only that axis anchored at the cursor; drag along pans it; drag across box-zooms the marked span. Pure gesture scope over the PR #117 clamp pipeline — `pan_axes`/`zoom_axes`/`zoom_limits`/containment/bounds govern unchanged.
- #111 — **structured hover + framework tooltips**: `xy:hover` gains an additive payload (`cursor.px`, `cursor.data` keyed by exact axis ID, resolved `points` with axis bindings and series color); `reflex_xy` grows a new `on_hover` prop (legacy `on_point_hover(row)` untouched); `xy.tooltip(render=…)` finally mounts — the client positions the component with the built-in placement logic, static charts included.

## Implementation spine

- **One mutation path** — `_applyStatePatch` in `js/src/57_viewstate.js` is the single "apply a state patch" implementation; `state_patch`/`view_nav`/`selection_rows` kernel messages, history restores, the static `root.xy` handle, and gestures all commit through the same validate → clamp → commit → emit path. Programmatic writes emit `source: "api"`, history emits `source: "history"`, so state bridges break echo loops by filtering.
- **Browser client owns live state** — history and the JS handle (`root.xy.applyState/state/back/forward`) work in kernel-less standalone HTML. Kernel-side `view_state()` is an event-fed cache: end-phase view events now always ship (one rAF-coalesced message per gesture); update-phase streams stay gated on listener presence.
- **Rows-selections are non-durable** (§5.1): kernel-resolved into the same mask buffers the gesture path ships, never in history, reported only as the opaque `{"rows": true}` marker.
- **v:1 strictness** — unknown keys/axes and non-finite numbers reject the whole document on both sides of the wire, never a partial apply.

## Validation (the §9 testing contract)

- `tests/test_view_state.py` — 21 kernel-side tests: §2 boundary rules, patch semantics, `view_nav`/`selection_rows` shapes, the `view_state()` cache lifecycle, brush-before-select ordering for programmatic geometric selects, the history switch, and the `Chart.select` attribute/method collision (caught by `ty`, fixed by privatizing the stored switch).
- `tests/test_view_state_client.py` — 9 headless-Chromium probes driving the real client: exact-f64 round-trip, clamp equivalence (programmatic vs gesture), history contract (N gestures → N entries, wheel coalescing, back/forward exactness, reset-then-back, linked pushes nothing, per-call opt-out, capacity 64), rows non-durability, axis-band scoping (wheel + drag mutate only the hovered axis; excluded axes get no band), additive hover payload beside the compatible legacy row, kernel-message application with `source: "api"`, and the custom tooltip mount.
- `tests/reflex_adapter/test_view_state_push.py` — 6 tests for the §5.2 push path (message shapes, no version bump, caller-thread validation, unwired behavior, module-level wrappers).
- Full suite: 2165 passed locally (plus reflex-adapter suite under a reflex venv: 77 passed; `test_socket_data_plane` failures reproduce on a pristine checkout in that ad-hoc venv — environmental, untouched code paths).
- `scripts/render_smoke_nonumpy.py` and `scripts/reflex_lifecycle_smoke.py` — clean (the lifecycle smoke exercises the restructured `XYChart.jsx` mount).
- `uv run ruff check .` / `ruff format --check .` / `pre-commit run --all-files` — clean; `ty check` clean on touched files.

Closes #110, closes #111, closes #120, closes #121.

### CODE REVIEW

File: spec/design/view-state.md [hunk 1]
```diff
@@ -0,0 +1,340 @@
...
+Rules:
+
+- `v` is the state-schema version, independent of the wire-protocol
+  handshake. A client that receives a higher `v` than it understands rejects
+  the whole document; it never applies a partial interpretation.
+- `ranges` is keyed by exact axis ID (`"x"`, `"y"`, `"y2"`, …), matching the
+  PR #117 event payload. Values are `[lo, hi]` f64 pairs.
+- `selection` is one of `{"range": {x0, x1, y0, y1}}`,
+  `{"polygon": [[x, y], …]}`, or `null`. The two non-null shapes are exactly
+  the shipped `on_brush` payloads.
```
Alek99: **Row selections need a durable representation.** The schema permits only range, polygon, or `null`, but §5 exposes `select(rows=...)` with history enabled. An arbitrary per-trace row selection cannot be serialized into either geometric shape, so `view_state()`, history restoration, and the round-trip invariant cannot preserve it. Please add a row-index/mask representation to the durable schema, or explicitly make row selections non-durable and non-undoable.

spec/design/view-state.md
```diff
@@ -0,0 +1,340 @@
...
+reflex_xy.reset_view(token)
+reflex_xy.select(token, range=..., polygon=..., rows=...)
+reflex_xy.clear_selection(token)
+reflex_xy.view_back(token); reflex_xy.view_forward(token)
+```
+
+Path: figure lock → one wire message (§8) → pushed room-wide as a `msg`
+event on the `/_xy` namespace → every client in the room applies it through
+the §3 mutation path with `source: "api"`. Multi-client semantics are
+therefore identical to `append` and `registry.publish`: the room converges.
```
Alek99, line 221: **Room-wide history navigation can diverge.** `view_back()`/`view_forward()` broadcasts a navigation operation to every client, but §4 makes history client-local. Clients can have different stacks after local gestures, so each can restore a different snapshot; even a partial `state_patch` can preserve different untouched axes. Please broadcast the resolved resulting state from one authority, or define navigation as mount/client-scoped instead of claiming room convergence.

Alek99 (commented):
The overall direction is strong, but these four contract gaps should be resolved before implementation so the durability and multi-client invariants remain achievable.

spec/design/view-state.md
```diff
@@ -0,0 +1,340 @@
...
+- `select(rows=...)` accepts per-trace row indices — the same canonical form
+  the shipped `Selection` object reports — and resolves them kernel-side
+  into the existing binary selection-mask buffers. `range=`/`polygon=` ship
+  the geometric forms and let the client resolve, exactly like a gesture.
+- `view_state()` does not round-trip: the kernel already observes every
+  committed change through view/selection events; it caches the last
+  durable state and serves reads from that cache. Immediately after a
+  write, the cache reflects the *requested* state only once the client's
+  event confirms the clamped result — reads are eventually consistent and
+  documented as such.
```
Alek99, line 192: **The kernel cache cannot observe drag-action changes.** The durable document includes `drag_action`, but the existing view and selection events do not carry it, and §8 explicitly adds no client→kernel message. A user changing the modebar tool therefore leaves the cache stale, so `view_state()` cannot return the last committed durable state. This needs a durable-state event/query path, or `drag_action` should be excluded from the cached API.

spec/design/view-state.md
```diff
@@ -0,0 +1,340 @@
...
+  "points": [
+    {"trace": "sensor A", "index": 1042, "row": {"x": 1731.0, "y": 0.81},
+     "color": "rgb(31, 119, 180)"}
+  ]
+}
+```
+
+`cursor.px` is chart-root-relative pixels (what a framework needs to
+position an element); `cursor.data` is the cursor position in data
+coordinates; `points` carries the resolved rows with series metadata. The
```
Alek99: **Cursor data is ambiguous with secondary axes.** A chart-root pixel maps to a different data value for each declared axis, so `cursor.data` containing only `x`/`y` is insufficient for traces bound to `x2`/`y2`. Please key cursor coordinates by exact axis ID and/or include each point's `x_axis`/`y_axis` so custom tooltips can interpret the coordinates correctly.

spec/design/view-state.md
```diff
@@ -0,0 +1,340 @@
...
+  documented as such.
+
+### 5.2 `reflex_xy` out-of-band API
+
+Mirrors `reflex_xy.append(token, …)` exactly — callable from any event
+handler, background task, or thread, with no component re-render and no
+payload reship:
+
+```python
+reflex_xy.set_view(token, ranges={"x": (t0, t1)})
```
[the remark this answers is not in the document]
masenf: this is actually okay. a view update is considered relatively cheap

python/reflex-xy/reflex_xy/assets/XYChart.jsx
```diff
@@ -93,43 +93,88 @@ export function XYChart(props) {
...
+    const el = outerRef.current;
     if (!el) return undefined;
     const start = (event) => cbRef.current.onAnimationStart?.(event.detail);
     const end = (event) => cbRef.current.onAnimationEnd?.(event.detail);
+    // view-state.md §7.1: the full payload rides the client's hover/leave
+    // events (bubbling CustomEvents), resolved without any kernel.
+    const hover = (event) => {
+      const d = event.detail || {};
+      if (d.cursor) {
+        cbRef.current.onHover?.({
```
Alek99: **[P1] Feed the structured payload into the custom renderer.** This handler forwards `active`/`cursor`/`points` only to `onHover`; the mounted `children` tooltip is never cloned or rendered with those values. As a result, `xy.tooltip(render=...)` displays static content unless the user builds a separate backend state bridge, and it cannot provide the advertised static/export behavior. Please use a client-side render-prop/portal contract (or otherwise pass this payload directly to the renderer).

js/src/57_viewstate.js
```diff
@@ -0,0 +1,548 @@
...
+      : null;
+    this._resetView(true, "reset", override);
+  },
+
+  // Kernel-resolved rows-selection (§5.1/§8): the same mask buffers the
+  // selection reply ships, applied as a NON-durable selection — never in
+  // history, reported in state only as the opaque marker.
+  _applyRowsSelection(msg, buffers) {
+    this._clearLassoOverlay();
+    this._applySelectionBuffers(msg, buffers);
```
Alek99, line 241: **[P1] Clear masks for traces omitted from a rows selection.** `_applySelectionBuffers` only replaces masks for entries present in `msg.traces`. After selecting rows on multiple traces, a later `select(rows={0: ...})` leaves the old masks active on every omitted trace while `_selectionCount` reports only the new rows. Clear all existing masks before applying a `selection_rows` document, or serialize empty entries for every selectable trace.

js/src/53_interaction.js
```diff
@@ -1398,6 +1433,13 @@ Object.assign(ChartView.prototype, {
       return before[0] !== after[0] || before[1] !== after[1];
     });
     if (!changedAxes.length) return [];
+    // History push (view-state.md §4): snapshot the pre-mutation durable
+    // state once per interaction id; linked/history sources never push.
+    this._historyRecord({
```
Alek99: **[P2] Do not record an intermediate snapshot when an animation is retargeted.** A second animated mutation records history here and then, at the existing `_viewAnim` branch below, merely replaces the in-flight target. Two rapid animated writes therefore store duplicate/intermediate snapshots instead of the first requested destination; the final event also retains the first call's source and interaction ID. Please settle/cancel the prior animation or update history and event metadata when retargeting.

python/xy/_figure.py
```diff
@@ -1391,6 +1405,155 @@ def append(
...
+        if not isinstance(rows, dict):
+            rows = {0: rows}
+        traces: list[dict[str, Any]] = []
+        out: list[bytes] = []
+        total = 0
+        for trace_id, indices in rows.items():
+            tid = int(trace_id)
+            if not 0 <= tid < len(self.traces):
+                raise ValueError(f"unknown trace id {trace_id!r}")
+            idx = np.asarray(indices, dtype=np.int64).ravel()
```
Alek99: **[P2] Validate canonical row indices before encoding them as `uint32`.** Negative, non-integral, duplicate, and out-of-range values are currently accepted. For a three-row trace, `rows=[-1, 99]` produces `[4294967295, 99]`; the browser selects nothing but reports `total: 2`. Reject invalid indices and compute totals from the validated/shipped rows.

Alek99 (commented):
The overall direction is strong, but these four implementation gaps should be fixed before merge: the custom tooltip renderer is not fed hover data, rows selections retain stale masks on omitted traces, retargeted animations corrupt history, and row indices are encoded without validation.

Alek99:
@greptile

PR #144, opened 2026-07-21
Skip heavy CI for docs-only PRs

Alek99:
## Summary

- skip the full CI workflow when a pull request changes only files under `docs/**` and/or `spec/**`
- apply the same exclusion to CodSpeed benchmarks
- keep mixed code and documentation pull requests on the normal CI path
- enforce both path exclusions in the workflow verifier and its regression tests

## Why

Documentation and design-spec pull requests currently trigger the full Rust, Python, JavaScript, browser, wheel, and benchmark matrix even though they cannot affect those artifacts. This avoids that unnecessary runner load while leaving the dedicated Docs workflow available for documentation changes.

## Validation

- `python3 scripts/verify_ci_workflow.py`
- `.venv/bin/python -m pytest -q tests/test_verify_ci_workflow.py` (49 passed)
- `uvx ruff check scripts/verify_ci_workflow.py tests/test_verify_ci_workflow.py`
- `uvx ruff format --check scripts/verify_ci_workflow.py tests/test_verify_ci_workflow.py`
- YAML parsing for both edited workflows
- `git diff --check`

GitHub-managed default CodeQL is configured separately and is not changed by this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,9 @@ on:
   push:
     branches: ["main", "claude/**"]
   pull_request:
+    paths-ignore:
+      - "docs/**"
+      - "spec/**"
   workflow_dispatch:
 
 # Multiple sessions push to two branches in bursts; without this the runner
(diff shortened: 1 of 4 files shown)


masenf (dismissed):
i think the tests added are kinda dumb, but sure

PR #155, opened 2026-07-22
Out-of-core canonical columns (disk-backed mmap store)

adhami3310:
## What

Realizes the **"mmap (native)"** canonical-store row of the memory model (dossier §27): a canonical column may be backed by a disk `np.memmap` instead of RAM, so a figure's **resident memory stays screen-bounded** (the density pyramid + grids), never data-bounded. This lets a scatter's canonical `x`/`y` exceed RAM — the target §2 has always listed as "100M+ / out-of-core" but never realized on native.

## How

The load-bearing observation: a NumPy `memmap` is a **transparent `ndarray`**. A column backed by one:

- satisfies the `ColumnStore` dedup key (`id(base)`, `data_ptr`, `nbytes`),
- passes straight to the ctypes kernels via the raw buffer pointer (`arr.ctypes.data`), so the OS pages it in on demand — `zone_maps`, `bin_2d`, `range_indices`, and the density pyramid all consume it **with no special-casing**,
- is dropped by the kernel as a rebuildable cache exactly like an in-RAM one.

So the store, zone maps, and every LOD tier already work out-of-core. The one operation that genuinely needs a new path is *creating* an array too large for RAM:

- `xy._ooc.MemmapF64Builder` streams canonical f64 to a disk memmap one batch at a time (peak RAM = one batch); `open_f64` reopens a column from disk.
- Feed the finished view to the ordinary API — `fig.scatter(x=col, y=col, density=True)` — and the whole pipeline runs against disk-backed truth.

## Honest accounting (§27)

`memory_report()` now separates:
- `canonical_bytes` — RAM-resident canonical (unchanged for all-RAM figures)
- `canonical_mapped_bytes` — disk-backed, OS-paged, reclaimable

An all-RAM figure reports `canonical_mapped_bytes == 0`, so nothing changes for existing usage.

## Tests

`tests/test_ooc.py` (6 tests): builder round-trip + growth, empty builder, ingest-without-RAM-copy, in-RAM report unchanged, zone-maps match RAM, and screen-bounded density first-paint over a memmap-backed scatter. Full suite green; ruff / ty (new files) / pre-commit clean.

## Spec

- §2 — out-of-core target marked realized
- §27 — new rule 5 (canonical may be out-of-core, native mmap)
- `lod-architecture.md` §4.4 — canonical out-of-core landed


adhami3310:
Addressed the review feedback in `3f9e162`:

**Greptile P1 — empty column reopened as a phantom `0.0` row (real bug, fixed).** `finalize()` left an 8-byte placeholder so `np.memmap` could map the empty column, but `open_f64()` derives the row count from file size, so the file reopened as one bogus `0.0` row. A memmap can't map an empty file, so an empty column is now a genuine **0-byte** file and both `finalize()`/`open_f64()` return a plain empty array. 0 bytes round-trips to 0 rows; a real single `0.0` row (8 bytes) stays one row. `test_empty_builder_roundtrips_to_zero_rows` pins both directions.

**CodSpeed −13.23% on `test_select_lasso_message_1m` — measurement artifact, not a real regression.** The benchmark's timed path is `select_polygon → select_range → ray-cast → wire mask`. Every file on that path is **byte-identical to main** on this branch (`channel.py`, `lod.py`, `channels.py` = 0 diff lines; `select_range`/`select_polygon`/selection-reply unchanged in `interaction.py`/`_figure.py`/`_payload.py`). Simulation-mode instruction counts are sensitive to allocation addresses, which shift when unrelated module-level code is added elsewhere in `interaction.py`/`columns.py` — so the delta is layout noise on unchanged code. Will acknowledge on CodSpeed.

PR #161, opened 2026-07-22
doc updates for seo and clarity

tgberkeley:
This PR updates the docs to make them more readable and more SEO ready


Alek99 (changes requested):
- The FAQ and related charts Should come below api ref
- Need to split up the sidebar
- Why are triangle mesh and facets and layers in charts they are more like componentss 

- Overall there is only one example per charts they need to be way more comprehensive

PR #177, opened 2026-07-22
Type the render client: no implicit any, no implicit this

Alek99:
Stacked on #159 (base: `feat/ts-vite-client`). Merge that first.

## What

Turns the TypeScript conversion from "compiles" into "typechecked". **2321 diagnostics → 0** under the newly enabled flags:

```
noImplicitAny  noImplicitThis  strictFunctionTypes  strictBindCallApply
noImplicitReturns  noFallthroughCasesInSwitch  useUnknownInCatchVariables
```

Every parameter, local, and `this` in the client is now typed — no value acquires `any` implicitly. `tsc` already gates every build (`js/build.mjs` runs it first, since oxc strips types without checking them), so this is enforced on every PR.

## How

- **`js/src/05_types.ts`** (new) carries the wire contract the kernel emits: `ChartSpec`, `TraceSpec`, `AxisSpec`, `ChannelSpec`, `ColumnMeta`, `GpuTrace`, `Hit`, `Row`, `Comm`, `TickResult`, `Theme`, `MarkKind`, `Rgba`. Deliberately permissive where the wire is permissive — open index signatures stay, because tightening one is a compatibility decision to make alongside the Python that emits it.
- **`as ThisType<ChartView>`** on the prototype-augmentation literals (51–57) types `this` as ChartView instead of the object literal. That one type assertion resolved **1190** of the 2321 diagnostics, and erases at compile.
- Per-module types where no shared name fits: LOD drill/density updates, animation easings and transition records, view-state patches and durable state, interaction option bags, GL program/buffer tags (`XyProgram`, `XyBuffer`), `ColumnView`, `AxisMap`.
- `_listen` became generic over its event type so call sites keep precise handlers (`PointerEvent`, `WheelEvent`, …) under `strictFunctionTypes`.

## Types-only — proven, not asserted

The committed minified bundles in `python/xy/static/` are **byte-identical** before and after this change; the diff contains no bundle. Independently, three of the workers transpiled HEAD vs. working copies and byte-compared the emitted JS per file.

That check earned its keep twice: it caught a `&&` → `?.` rewrite that had slipped into a catch-narrowing fix, and a case where an interface placed above the first runtime statement absorbed a file's header comment and changed emit.

**Verification:** `tsc` clean · bundles byte-identical · full pytest **2208 passed** · render smoke `XY_OK` with pixel counts and context-loss hashes identical to the pre-typing run · ruff + pre-commit clean.

## Test fallout

Source-marker tests that grep exact client signatures are updated to the typed forms, preserving their parameter-contract intent. Separately, `tests/reflex_adapter/test_assets.py` still grepped *unminified* markers (`function renderStandalone(`) — it only runs when reflex is installed, so it was missed when the bundles began shipping minified in #159. Fixed here to use the same export-alias contract `verify_wheel.py` asserts.

## Latent bugs surfaced (reported, not fixed — types-only)

Typing the code found real issues. None are touched here; each deserves its own change:

1. `lodApplyDensityUpdate` reads `g.density.colormap/.color/.lut` unguarded, while every sibling access guards — a density update on a trace whose density was never built is an uncaught TypeError.
2. `RECT_MARK`/`BAR_MARK`/`SEGMENT_MARK`/`AREA_MARK`/`line` dereference `g.trace.style.color` unguarded; `style` is optional on the wire (`MESH_MARK` and hexbin do guard).
3. `54_kernel` builds `pendingTraceIds` with `Number(upd.id)` but tests membership with the raw `g.trace.id`; `TraceSpec.id` is `string | number`, so a string id silently skips the `clearPending` sweep.
4. `_onA11yKey` computes `hit.trace * 1e9 + hit.index` from a trace *id*; a string id makes `_hoverId` NaN and every `!== -1` guard read true forever.
5. `51_annotations` `liftBounds.width` is only safe via a two-expression short-circuit that a future edit could split.
6. `57_viewstate` `canBandPan()` returns truthy via `_axisContained` even when pan policy excludes the axis — possibly deliberate, worth a look.

Items 1, 2, and 5 are exactly the class `strictNullChecks` would catch, which motivates the order below.

## Deferred (documented in `js/tsconfig.json`)

1. Remove `ChartView`'s `[key: string]: any`. It is what still lets `this.typo` compile and what stops the annotated signatures in 51–57 from being enforced at call sites. Needs ~403 declared members (172 class methods, 119 augmentation methods via `declare module` merging, 112 instance fields) — mechanical, but its own reviewable PR.
2. `strictNullChecks` — largest remaining win, largest diff.
3. `strict: true` once 1 and 2 land.

### DIFF
--- a/js/src/00_header.ts
+++ b/js/src/00_header.ts
@@ -37,7 +37,25 @@ const XY_FRAME_DEFAULT_LIMITS = Object.freeze({
   maxBufferBytes: 256 * 1024 * 1024,
 });
 
-export function xyByteSpan(value, label = "buffer") {
+/** Per-field caps on an inbound frame; callers may override any subset. */
+export interface FrameLimits {
+  maxFrameBytes?: number;
+  maxMetadataBytes?: number;
+  maxBuffers?: number;
+  maxBufferBytes?: number;
+}
+
+/** A decoded XYBF frame: the JSON metadata message plus zero-copy buffer
+ * spans into the caller's ArrayBuffer, and the envelope facts a caller may
+ * want to log or assert on. */
+export interface DecodedFrame {
+  message: any;
+  buffers: Uint8Array[];
+  version: number;
+  byteLength: number;
+}
+
+export function xyByteSpan(value: unknown, label = "buffer"): Uint8Array {
   if (value instanceof ArrayBuffer) return new Uint8Array(value);
   if (ArrayBuffer.isView(value)) {
     return new Uint8Array(value.buffer, value.byteOffset, value.byteLength);
@@ -45,7 +63,7 @@ export function xyByteSpan(value, label = "buffer") {
   throw new TypeError(`${label} must be an ArrayBuffer or ArrayBuffer view`);
 }
 
-export function bytesToSpan(b) {
+export function bytesToSpan(b: unknown): Uint8Array {
   const span = xyByteSpan(b, "chart payload");
   // anywidget/third-party callers may hand us an oddly-offset DataView. Keep
   // the normal aligned path zero-copy; preserve compatibility with one narrow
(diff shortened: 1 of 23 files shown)


Alek99:
Docs verification passed on `99cadae3cfbbc59de0b63ce09145351154600259`.

- `docs/app` suite: **72 passed, 1 expected xfail**
- Production docs build completed successfully.
- Sitemap, Markdown-asset, and generated-HTML route validators passed.
- All **57 published `/docs/xy` sitemap routes** returned HTTP 200 from the production server.
- Browser smoke test rendered the landing page’s interactive charts (12 canvases) with no console errors.

The PR changes only the typed render client; the documentation site remains healthy.

PR #198, opened 2026-07-22
ENG-10514: Update toolbar style

carlosabadia:


Alek99 (changes requested):
Needs to address the changes in slack

PR #201, opened 2026-07-22
fix(reflex): stream on_view_change during gestures with a latest-wins throttle

FarhanAliRaza:
## Problem

PR #113 replaced the Reflex wrapper's immediate `on_view_change` dispatch with a **200 ms trailing-edge debounce**. The render client emits a `view_change` per animation frame during a pan/zoom, so every frame reset the timer — the handler could only fire once the gesture went quiet. An `on_view_change`-computed detail chart (showcase §4) froze during the interaction and only re-rendered ~a second after it ended, where it previously tracked the gesture in real time.

## Fix

`dispatchView` in `XYChart.jsx` now uses a **leading + trailing, latest-wins throttle** (`VIEW_THROTTLE_MS = 120`, matching the hover throttle):

- the first event of a gesture dispatches immediately,
- updates stream at most once per 120 ms while the gesture is in progress,
- a trailing flush guarantees the resting viewport always lands as the last event.

The envelope's `phase` field now reflects the client's gesture phase — `"update"` mid-gesture, `"final"` at rest — so handlers that only care about settled views can filter on it. `linked`/`republish` source suppression and unmount timer cleanup are unchanged. Handlers written against the old contract keep working: they simply receive intermediate `update` events before the same `final` one.

## Verification

- Drove the running showcase app headlessly (repo CDP driver): 24 wheel-zoom steps at 140 ms spacing — a continuous gesture the old debounce can never fire inside — produced **23 mid-gesture Reflex state updates**, detail histogram tracking each step, with the exact resting view (`phase: "final"`) arriving last.
- `scripts/reflex_ws_smoke.py` mount/shared-websocket/pixel assertions pass. Its §16 density-drill step times out, but it fails identically with the unmodified main `XYChart.jsx` swapped into the served app — pre-existing on main, unrelated to this change (tracked separately).
- `pre-commit run --all-files`, `ruff check`, `ruff format --check` all green; the wrapper contract test assertions replayed against the modified JSX all hold.

## Spec

Per the spec-first contract: `spec/design/reflex-integration.md` now documents the throttle + phase stream (and drops a stale line-number cite), `ViewChangeEvent` typing/docstring, `python/reflex-xy/README.md`, the demo comment, and `tests/reflex_adapter/test_component.py` are updated to pin the new behavior.

### DIFF
--- a/examples/reflex/xy_reflex_demo/xy_reflex_demo.py
+++ b/examples/reflex/xy_reflex_demo/xy_reflex_demo.py
@@ -262,8 +262,9 @@ class Demo(rx.State):
     @rx.event
     def on_view(self, event: reflex_xy.ViewChangeEvent):
         # `event` is the v1 view-change envelope; `x_domain` is the reported
-        # [x0, x1] window (debounced by the wrapper). Store the window; the
-        # `detail` figure var depends on it and recomputes.
+        # [x0, x1] window (throttled by the wrapper, streaming during the
+        # gesture). Store the window; the `detail` figure var depends on it
+        # and recomputes.
         x_domain = event.get("x_domain") or [0.0, 0.0]
         self.view_x0 = float(x_domain[0])
         self.view_x1 = float(x_domain[1])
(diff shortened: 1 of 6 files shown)


masenf (commented):
do the reflex built-in throttle and debounce event actions not work with this event? if the event debouncing is only in the reflex wrapper, do we need to re-implement this functionality in a way that is not configurable by the developer?

masenf (approved):

PR #217, opened 2026-07-22
Area-weight pyramid compose to remove interim zoom banding (#153)

### LINKED ISSUE #153: vertical "banding" when zooming in through LoD chart
when zooming through a very dense point cloud sometimes the interim view while loading the drill data shows vertical banding.

adhami3310:
Fixes #153.

## Diagnosis

The vertical banding is a **resampling beat (moiré) artifact** in the tile-pyramid `compose` path — visible in the interim density frames served while zooming a dense point cloud.

The client requests a density grid at ~screen resolution (`w`×`h`), and `compose` picks *the coarsest pyramid level that still meets that resolution*. By construction that level packs **between `w` and `2w` source cells** across the window (source→output ratio in `[1, 2)`). The old code assigned each source cell **whole** to the single output bin under its center:

```rust
out_row[ox as usize] += c as f32;   // ox = floor of the cell center
```

At a ratio like 1.33, adjacent output bins receive **1 vs 2** source cells apiece — a ~2:1 brightness beat that renders as the regular vertical stripes in the issue screenshot (dominant on X because the canvas is wider than tall, so the X ratio lands in the beat zone).

## Fix

Replace the center-only assignment with **area-weighted resampling** (a box downsample): each source cell splits its count across the output bins its extent overlaps, in proportion to overlap. A uniform field now composes flat instead of beating. A new `axis_weights` helper precomputes the per-axis split (hoisted, so `compose` stays O(visible cells)).

**Exactness preserved**: weights within `COMPOSE_SNAP_EPS` (1e-6) of a bin edge collapse to a single bin, so cell-aligned windows (ratio exactly 1) remain **bit-exact** against `bin_2d` — the documented contract and existing tests hold.

## Verification

- All 9 `tiles` Rust tests pass, including the two exact-alignment tests and conservation.
- New regression test `compose_unaligned_ratio_has_no_banding` — confirmed it **fails** under the old nearest-neighbor logic (column = 128 vs mean 85.3) and passes with the fix.
- Full `cargo test` (103 tests), release build, ABI smoke (121 checks), clippy, fmt all clean.
- End-to-end through the Python binding on the built library: full-window compose still bit-exact; uniform field composes flat (max column deviation 0.000 vs the prior 2× swing).

## Files changed

- `src/tiles.rs` — area-weighted `compose` + `axis_weights` helper + regression test + module doc.
- `spec/design/lod-architecture.md` — disclosed the resampling method (§28 "no silent decisions").

### DIFF
--- a/spec/design/lod-architecture.md
+++ b/spec/design/lod-architecture.md
@@ -374,7 +374,14 @@ contract entry before it lands.
    drill threshold; `compose` picks the coarsest level that still meets the
    render resolution and refuses beyond `MAX_UPSAMPLE` (2×), so below-floor
    and near-drill windows fall through to the exact `range_indices` +
-   `bin_2d` path. Level is recorded per update as `binning: "pyramid-L<l>"`.
+   `bin_2d` path. The coarsest adequate level packs 1–2 source cells per
+   output bin, so `compose` **area-weights** each source cell across the bins
+   its extent overlaps rather than assigning it to the bin under its center;
+   center-only assignment handed adjacent bins 1 vs 2 cells apiece — a beat
+   against the output grid that showed as vertical banding in interim aggregate
+   frames while zooming a dense cloud (#153). Weights within a snap tolerance of
+   a bin edge collapse to one bin, so cell-aligned windows stay bit-exact.
+   Level is recorded per update as `binning: "pyramid-L<l>"`.
    Traces on a nonlinear (log/symlog) axis skip the pyramid entirely — its
    raw-space levels cannot compose a scale-coordinate grid (dossier §28) —
    and always take the exact path.
--- a/src/tiles.rs
+++ b/src/tiles.rs
@@ -7,10 +7,12 @@
 //! conserves total count bit-exactly.
 //!
 //! `compose` fills a render grid for a view window from the coarsest level
-//! that still meets the render resolution (bounded upsampling), assigning
-//! each source cell to the output bin containing its center — deterministic
-//! and count-conserving over whole cells; windows that outresolve level 0
-//! are refused (caller falls back to an exact re-bin, disclosed per §28).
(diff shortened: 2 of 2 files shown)


adhami3310:
### Re: the CodSpeed `test_pyramid_compose` regression

Flagging this as an **intentional, understood tradeoff** — the acknowledge action is left to a maintainer.

**What the number is.** CodSpeed runs this in **Simulation** mode (cachegrind-style), which scores a *cost model of instructions executed*, not wall-clock. It doesn't model the out-of-order execution / memory-level parallelism that dominate this kernel on real silicon.

**Why it moved.** This PR is the #153 banding fix: it replaces `main`'s center-only (nearest) binning with **area-weighted resampling**. Area weighting inherently does ~2× the arithmetic per source cell (each cell splits across up to two bins per axis, with float weights) — so under an instruction-count model, ~2× is unavoidable. It's the cost of not lying about density at interim zoom levels.

**Real wall-clock is at parity or better.** The baseline is memory-bound, so the extra arithmetic hides behind memory stalls. Measured locally on the exact CodSpeed shape (2.1M points, 512×384 served from the 1024 level):

| version | wall-clock |
| --- | --- |
| `main` (nearest) baseline | 0.919 ms |
| one-pass 2D area-weight scatter (first cut) | 2.22 ms |
| separable pull-based (current HEAD, e73487b) | **0.950 ms** |

`e73487b` makes the area-weighted compose essentially free relative to the nearest-neighbor baseline it "regressed" from (@alekpetuskey measured 0.53 ms vs 0.63 ms on his machine — faster than baseline). All 10 `tiles` tests pass and the uniform-field column deviation is 0.0 (banding gone).

**Recommendation:** acknowledge on CodSpeed. Instruction count is genuinely higher by design; there's no way to make area weighting as cheap as nearest without reintroducing the banding, and on-hardware latency is unchanged.

PR #222, opened 2026-07-23
Reject raw-encode domains that collapse after f32 conversion

Alek99:
Review fixes for #187's `_raw_wire_ok` guard, split out of #221 when that PR retargeted to main (these fixes belong to the raw-encode code, which lives on this stack).

- **P1** — the guard accepted domains whose f64 span survives an f32 cast but whose *endpoints* collapse to the same f32 (`[1e30, 1e30 + 1e20]`): every shipped value rounds to `f32(lo)` and the shader maps them all to the domain floor. It also accepted spans whose f32 reciprocal overflows (`[0, 1e-45]`), sending `Inf` into the map uniform. Both now require `float32(lo) < float32(hi)` and a finite f32 inverse span, falling back to the f64-normalized unit encode; tests pin both cases.
- **P2** — the reciprocal bound is compared in f64 against `np.finfo(np.float32).max`, never cast through `np.float32`: the demoting cast warns on overflow for exactly the subnormal spans being rejected, and `np.seterr(over="raise")` setups would turn that into a `FloatingPointError` inside `build_payload`. The overflow test runs its probe under `np.errstate(over="raise")` to pin it.

Full `tests/test_scatter.py` passes (82); the fallback tests also pass with `-W error::RuntimeWarning` (no warnings emitted).

### DIFF
--- a/python/xy/channels.py
+++ b/python/xy/channels.py
@@ -507,13 +507,16 @@ _F32_ABS_MAX = 3.0e38
 def _raw_wire_ok(domain: tuple[float, float]) -> bool:
     lo, hi = float(domain[0]), float(domain[1])
     span = hi - lo
-    return (
-        abs(lo) < _F32_ABS_MAX
-        and abs(hi) < _F32_ABS_MAX
-        and np.isfinite(span)
-        and span > 0.0
-        and float(np.float32(span)) > 0.0
-    )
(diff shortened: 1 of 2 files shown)


Alek99:
Folding back per review direction — keeping #221 as the single PR. These two fixes patch `_raw_wire_ok`, which only exists on the #186–188 stack (not on main, where #221 now bases), so they cannot ride in #221's diff. They remain on branch `alek/raw-wire-f32-guard` (commits 1daca36, acc16f4), ready to cherry-pick into the stack or onto main when the stack rolls up.

PR #226, opened 2026-07-23
Density surfaces wear mean point color (LOD doc §2)

masenf:
Tier-2 scatter aggregation now colors each density cell with the alpha-weighted mean of the point colors in that cell, averaged in linear light. The count still drives the alpha channel for opacity/coverage. This eliminates the visual lie of colormapping density counts and lets zoomed-out views truthfully represent the data's own colors.

## Key changes

**Rust core (`src/`):**
- Added `bin_2d_mean_color_cells` kernel that accumulates exact integer sums of alpha-weighted linear-light color alongside counts, enabling order-independent parallel reduction
- New `Pyramid::build_color` path that builds mean-color planes alongside count pyramids; `reduce_color_level` merges them during pyramid construction
- Static sRGB↔linear-light u16 lookup tables (`SRGB_TO_LINEAR_U16`, `linear_u16_to_srgb_u8`) for bitwise-deterministic color math independent of thread count
- Bumped `ABI_VERSION` to 40; added `xy_bin_2d_mean_color` and `xy_pyramid_build_color` C exports

**Python (`python/xy/`):**
- `interaction.py`: Channel-bearing traces now build mean-color pyramids; appends invalidate and rebuild (color domain changes require rescan)
- `channels.py`: New `quantize_unit_u8` and `colormap_lut_rgba8` helpers for wire quantization and LUT export
- `_payload.py`: Density updates now ship mean-color RGBA8 planes alongside counts; continuous channels quantize to u8 for wire efficiency
- `marks.py`: Color channel survives aggregation as the density surface's per-cell mean; size/density_val channels are dropped (count is implicit in alpha)

**JavaScript (`js/src/`):**
- `45_lod.ts`: `lodWriteGridTexture` now handles RGBA mean-color planes; premultiplies RGB by the log-tone-mapped alpha so bilinear filtering weights color by coverage
- `54_kernel.ts`: Zoom-in request elision (T12) — views inside an exact drill window skip the round-trip and use cached marks
- `50_chartview.ts`: Density traces no longer get a colormap gradient swatch in the legend (they wear the data's own colors, not a colormap)

**FastAPI example (`examples/fastapi/live_drilldown.py`):**
- Simplified from custom density binning to delegating all LOD logic to the engine's `Figure.density_view` and `Figure.pick`
- Removed `DensityOverview` class and integral-image math; the pyramid now lives in the engine
- Uses `encode_frame` from `xy.channel` for XYBF binary transport (no base64)
- Removed request bookkeeping; the render client handles stale-reply deduplication via `seq`

**Documentation & tests:**
- `spec/design/lod-architecture.md` §2: Documented mean-color aggregation law and per-channel rules
- New `tests/test_density_mean_color.py`: Kernel correctness vs NumPy oracle, wire shape, static exporter color law
- New `tests/test_drill_zoomin_elides_request.py`: T12 elision (no re-request inside exact drill window)
- Updated docstrings and examples to reflect the new density behavior

## Implementation notes

- Mean-color planes are optional (`color_levels: Option<Vec<Vec<[u16; 4]>>>` in `Pyramid`); count-only traces skip them
- Color accumulation is exact integer arithmetic (sums, not means) so parallel merges are order-independent
- Per-level rounding error is ≤ 0.5 lsb of u16 (recorded in LOD doc §28)
- Density surfaces no longer carry a separate `density_colormap` field; the surface IS the data's colors
- Zoom-in request elision (T12) bumps `seq` without sending a message, so stale replies from wider views are dropped correctly

https://claude.ai/code/session_01VYW3bncfneK8NbLoAFDMoo


masenf:
This PR seems to have a problem inflating the wire transfer size of the drilldown payload significantly. Needs to be fixed before merging.

PR #228, opened 2026-07-23
Re-land separable pyramid compose (recover #153 perf)

adhami3310:
## Re-land the separable pyramid compose (perf recovery)

Follow-up to #217. That PR shipped the #153 banding fix (area-weighted `compose`), but the **separable perf-recovery commit was dropped before merge**, so `main` currently runs the one-pass 2D-scatter form — 4 scattered read-modify-writes per source cell.

This restores the separable implementation (Alek's original commit, authorship preserved): resample each source row into a `w`-wide scratch row via **pull-based** per-bin taps transposed from `axis_weights`, then distribute that row into its 1–2 output rows with contiguous, vectorizable scale-and-add loops.

### Why it's a strict win

- **Same result.** Pure reindexing of the same weights, same per-axis summation order → the cell-aligned bit-exactness contract against `bin_2d` holds. All 10 `tiles` tests pass unchanged; the #153 banding stays fixed.
- **~2.3× faster wall-clock.** On the `test_pyramid_compose` shape (2.1M points, 512×384 from the 1024 level):

  | version | wall-clock |
  | --- | --- |
  | pre-#153 nearest baseline | 0.919 ms |
  | one-pass 2D scatter (**current main**) | 2.22 ms |
  | separable (**this PR**) | 0.968 ms |

  Back to baseline parity.
- **CodSpeed:** this compares against current `main` (which already has the one-pass area-weighting), and the separable form executes *fewer* instructions than the one-pass 2D scatter — so it should read as an improvement rather than the regression seen in #217 (where the base was pre-area-weighting `main`).

No API/ABI change; `compose` signature and all callers are untouched.

### DIFF
--- a/src/tiles.rs
+++ b/src/tiles.rs
@@ -269,28 +269,46 @@ pub fn compose(
         // within COMPOSE_SNAP_EPS of a bin edge collapse to one bin, keeping
         // cell-aligned windows exact. center_range can yield an empty range for
         // windows astronomically past the domain; the guard avoids a slice panic.
+        //
+        // The 2D split is separable (weight = wx·wy), so resample in two
+        // passes: gather each source row into a w-wide scratch row by the x
+        // weights, then distribute that row into its 1–2 output rows as
+        // contiguous scale-and-add loops the compiler vectorizes. The x pass
+        // is pull-based — per output bin, sum the 1–3 source taps transposed
+        // from axis_weights — because both push formulations lose: the fused
+        // 2D scatter costs 4 read-modify-writes per cell, and a 1D x scatter
+        // chains through memory (adjacent cells hit the same bin, stalling on
+        // store-forwarding); pulling makes every output bin independent, so
+        // the loop pipelines. Scaling by wy after the x sum also rounds once
+        // instead of per cell; the aligned case (single tap, weight exactly
+        // 1.0) remains bit-exact against bin_2d.
         let none = u32::MAX;
         let xw = axis_weights(cx0, cx1, p.x0, cell_x, lo_x, sx, w);
         let yw = axis_weights(cy0, cy1, p.y0, cell_y, lo_y, sy, h);
+        let (xstarts, xtaps) = transpose_to_taps(&xw, w);
+        let mut xrow = vec![0.0f32; w];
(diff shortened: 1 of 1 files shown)


adhami3310:
### CodSpeed status: recommend **acknowledge** — the metric can't see this optimization

Short version: this PR is **2.4× faster in real wall-clock** but CodSpeed (simulation/cachegrind mode) scores it ~10% *slower*, because the win is a CPU-pipeline effect a cachegrind cost model has no concept of.

**Measured on the `test_pyramid_compose` shape (2.1M pts, 512×384 from the 1024 level):**

| version | CodSpeed *simulated* | real wall-clock (local) |
| --- | --- | --- |
| pre-#153 nearest | 5.9 ms | ~0.63 ms |
| one-pass 2D scatter (**current main**) | 9.8 ms | 2.22 ms |
| separable (**this PR**) | 10.9 ms | **0.91 ms** |

**Why the metric is inverted here.** The one-pass form does `out[...] += …` where adjacent source cells land on the same output bin ~40% of the time (source→output ratio ≈1.6), creating store-to-load-forwarding dependency chains that serialize on real silicon. The separable pull sums each bin independently in a register, so it pipelines — hence 2.4× faster wall-clock. Cachegrind models instructions + cache but **not** store-forwarding latency, so it can't see the stall being removed; it only sees the pull structure's slightly higher instruction count and scores it worse.

**I tried to make it cheap on both metrics** (fused the y-distribute into the pull loop to drop the scratch-row buffer round-trip) — it didn't move the simulated number (stayed 10.9 ms) and was marginally slower in wall-clock, so I reverted it. The conclusion: area-weighting's simulated cost is irreducible relative to the one-pass; no code arrangement wins on a metric blind to the actual bottleneck.

**Note:** main already carries the one-pass area-weighting (#217), i.e. the 5.9→9.8 ms simulated jump was already accepted. This PR trades a further +11% simulated for a 2.4× real speedup.

**Recommendation:** acknowledge this benchmark on CodSpeed. Longer term, `test_pyramid_compose` is a candidate to move to CodSpeed **walltime** mode (the repo already segregates wall-clock-sensitive benches out of simulation) so it rewards the real optimization — happy to open that as a follow-up if we want it.

PR #230, opened 2026-07-23
Refresh README for adoption

Alek99:
## Summary

- add transparent XY binned-scatter artwork at the top of the README
- refresh positioning, installation, and docs links for Reflex-hosted docs
- add a positive `Is xy for me?` section and clarify current capabilities
- update the “How it works” architecture diagram to reflect the shipped data flow and representation bounds
- remove the License and contributor-setup sections from the README

## Validation

- rendered the artwork from the documentation’s canonical XY demo
- compiled the Reflex documentation app in dry mode
- ran `git diff --check` and the repository claim guardrails

PR #238, opened 2026-07-24
Grey, persistent box/lasso/x-range/y-range selections

Alek99:
## What

Two changes to how selections look and behave.

**1. Grey defaults matching the toolbar.** The box, lasso, x-range, and y-range overlays defaulted to a blue (`rgba(90,140,240,…)`) that matched nothing else in the chrome, and box-zoom used a separate flat grey. All of them now default to the modebar's own text greys via scheme-aware internal tokens `--xy-selection` / `--xy-selection-fill`:

- light: `rgba(92,101,115,.6)` stroke / `…,.12)` fill (`#5c6573`, the modebar text grey)
- dark (`.dark` on the chart root or any ancestor): `rgba(173,180,191,.6)` / `…,.12)` (`#adb4bf`)

The public `--chart-selection*` / `--chart-zoom-selection*` tokens are unchanged and still override both schemes; only the built-in fallbacks moved.

**2. Persistent box / x-range / y-range selections.** Previously only the lasso stayed drawn after the gesture — box and range brushes vanished the instant the drag ended, leaving just the highlighted points. Now all three persist like the lasso: the data-space rectangle re-projects onto the selection overlay every draw (through pan and zoom) until the selection is cleared.

- The overlay is shaped by a `mode` discriminator (box | x | y). Range brushes span the full plot on their free axis and drop the cross-axis border pair: **x-range keeps only its left/right edges, y-range only its top/bottom** — so each reads as the bounds it constrains.
- `mode` round-trips through durable state as an optional `selection.range.mode` (a plain box omits it, keeping the existing `{x0,x1,y0,y1}` shape) and never affects the selected point set.
- Box and lasso overlays stay mutually exclusive; double-click still clears; rows-selections clear both.

Spec and docs (`spec/api/styling.md`, `spec/design/view-state.md`, `docs/styling/*`) updated to match.

## Verification

- `node js/build.mjs` (tsc typecheck + bundle) passes
- New browser probe `test_rectangular_selection_persists_with_range_borders` covers persistence through the real draw path, the range-edge borders, `mode` round-trip via `applyState`, rejection of a bad `mode`, and box/lasso mutual exclusion
- 195 passed across `test_view_state_client`, `test_static_client_security`, `test_components`, `test_ui_issue_regressions`, `test_modebar_select_drill`, `test_selection_rows`, `test_view_state`
- Browser-verified on a standalone HTML export: box drag stays drawn and re-projects on zoom; x-range shows only vertical edges spanning full height; y-range shows only horizontal edges spanning full width; grey flips to the dark palette under a `.dark` root class

## Summary by CodeRabbit

- **New Features**
  - Rectangular, x-range, and y-range selections now remain visible after redraws, cancellations, and state restoration.
  - Selection overlays can be restored through linked charts and saved view state.
  - Lasso and rectangular selections now correctly replace one another.

- **Bug Fixes**
  - Improved clearing and restoration of selection overlays.
  - Added validation for selection range modes.

- **Style**
  - Selection and zoom overlays now use scheme-aware neutral colors, including dark mode.


Alek99:
The two red checks (`Python 3.11 floor`, `Test (Rust + Python + JS)`) are a **pre-existing `main` breakage**, not from this PR. All 20 failures are in `tests/test_docs_examples.py` / `tests/test_verify_local.py` asserting README prose and `make` shortcuts that #230 ("Refresh README for adoption", the tip of main) removed — main's own CI run at b6ab496 is red for the same reason. This PR's commits touch no README or doc-test files.

#236 ("Decouple CI tests from README", MERGEABLE) fixes exactly these. I merged #236 into this branch locally (no conflicts) and ran the previously-failing suites plus this PR's new selection test — **117 passed**, including all 20 that CI reports red. So once #236 lands and this branch picks it up, CI goes green with no change needed here.

Alek99:
Both Greptile findings addressed in be421f6:

**P1 — Restore linked overlays.** Fixed. The link-group handler in `js/src/50_chartview.ts` now hydrates the same persisted overlay state a local gesture produces: a linked `range` sets `_boxSelection` (with the band `mode`) and clears any lasso; a linked `polygon` sets `_lassoPolygon` and clears any box; `clear` wipes both. `_selectLocal`/`_selectLocalPolygon` already call `draw()`, so `_drawNow` now re-projects the band on the peer. I also made `_sendSelect` broadcast the range **with** its x/y `mode` so peers render the correct edge-only shape (the wire previously carried only `{x0,x1,y0,y1}`). This also fixes a latent bug where a linked selection left a stale opposite-type overlay drawn.

**P2 — Document selection overrides.** Fixed in `spec/api/styling.md`: the override sentence now lists `--chart-selection` / `--chart-selection-fill` and `--chart-zoom-selection` / `--chart-zoom-selection-fill` alongside the modebar/badge tokens, noting the CSS resolves each public token ahead of the internal `--xy-selection*` fallback. (Kept the current `.6` border alpha rather than the `.9` in the suggestion, which predated the later border-lightening commit.)

Added `test_linked_selection_hydrates_persisted_overlay`, which drives the real `BroadcastChannel` handler with the exact message shapes the sender emits (x-range, polygon, plain box, clear) and asserts overlay + durable state on the receiving peer. Also verified end-to-end in a browser with two link-grouped charts: an x-range brushed on one chart draws the grey vertical-edge band on the linked peer.

PR #240, opened 2026-07-24
Fix pyplot autoscale margins and bar extents [mpl compatibility]

sselvakumaran:
## Summary

- snapshot Matplotlib's default x/y autoscale margins when each axes is created, and apply them consistently to public limits and rendered domains
- compute bar autoscale extents from the full geometry: category width, zero or explicit baseline, and cumulative stacked endpoints
- keep bar baselines and contour boundaries sticky while adding Matplotlib-style headroom, including log-coordinate padding
- add regressions for line/log margins, positive/negative/mixed/stacked/horizontal/categorical bars, and bar-label headroom

## Validation

- `pytest -q tests/pyplot` — 623 passed, 3 skipped
- focused autoscale/layout/reference suite — 50 passed
- `ruff check` — passed
- `ruff format --check` — passed
- `git diff --check` — passed

The visual micro-suite also confirms line views now use the expected 5% limits and ordinary/stacked/horizontal bar labels remain inside the rendered view.


sselvakumaran:
## Before / after renderings

These are the same pyplot gallery cases at the same normalized 640×480 output size. **Before** is the pre-PR behavior; **after** is this PR.

### Horizontal bars — full value/category extents

Before, every bar is clipped to the fallback `0..1` x-domain. After, the data extent, category width, and label headroom are included.

| Before | After |
| --- | --- |
| Horizontal bars before | Horizontal bars after |

### Rotated categorical bars — all categories remain in view

Before, the fallback category domain shows only two of four bars. After, all four categories and the outer half-bar widths fit.

| Before | After |
| --- | --- |
| Rotated categorical bars before | Rotated categorical bars after |

### Positive bar labels — Matplotlib-style headroom

Before, the tallest value label crosses the top spine. After, the 5% autoscale margin keeps it inside the axes.

| Before | After |
| --- | --- |
| Positive bar labels before | Positive bar labels after |

The largest changes are in the plot limits and bar visibility. This PR does **not** fix the separate tick-label-to-tick spacing/chrome issue; that remains visible in both sets.

sselvakumaran:
Correction to the earlier visual: that panel compared two XY revisions, not Matplotlib directly. This one renders the same calls in Matplotlib 3.11 and XY:

Direct Matplotlib 3.11 versus XY comparison

What the Matplotlib source and direct measurements show:

- General autoscaling is a transformed data-span margin (`delta = span * margin`), with the default `axes.xmargin` / `axes.ymargin` set to 0.05. The automatic line example therefore has the same `0.475..1.025` limits in both implementations.
- `bar_label` creates an `Annotation` in offset points; its text bbox does not enter `dataLim`. In the exact Matplotlib reference above, the `1.8` bbox reaches about 0.46 px past the top spine.
- XY now keeps Matplotlib's general 5% autoscale behavior, but raises the automatic value-axis reservation to 7.5% when an edge `bar_label` is present. That gives the label visible headroom on vertical and horizontal bars. Explicit `set_xlim` / `set_ylim` and explicit `margins()` still take precedence.
- The default XY PNG is intentionally high-DPI (1280x960 for this 640x480 figure). The image above normalizes it to 640x480 only for same-size viewing; the 2x scale multiplies marks, text, and gaps uniformly and is not the source of the relative margin.

The separate tick-label fix is also included: padding starts after the outward end of the tick, and the browser geometry test measures the requested 5 px gap after a 6 px tick. Validation: 710 passed, 3 skipped, plus the focused browser geometry test.

PR #242, opened 2026-07-24
Fix GridSpec and aspect gallery APIs [mpl compatibility]

sselvakumaran:
## Summary

Fix two Matplotlib 3.11 gallery failures in the pyplot compatibility layer:

- accept the positional `adjustable` argument used by `Axes.set_aspect("equal", "box")`
- expose `Axes.get_gridspec()`, `get_subplotspec()`, and `Axes.remove()`
- preserve original subplot cell geometry when arbitrary axes are removed and replaced by a spanning GridSpec subplot

This is intentionally separate from #240 and does not change tick or default margin behavior.

## Gallery references

- [Equal axis aspect ratio](https://matplotlib.org/stable/gallery/subplots_axes_and_figures/axis_equal_demo.html)
- [Combine two subplots using subplots and GridSpec](https://matplotlib.org/stable/gallery/subplots_axes_and_figures/gridspec_and_subplots.html)

### Equal aspect — before / Matplotlib / after

Equal aspect comparison

Before, the example aborts at the fourth axes and leaves it elliptical. After, the positional API works and the fourth plot is a circle with the expected auto-adjusted limits.

### GridSpec spanning — before / Matplotlib / after

GridSpec comparison

Before, the example aborts with a 3×3 grid. After, the two lower-right cells are removed and replaced with the intended spanning axes and annotation.

## Verification

- `539 passed, 65 skipped` in `tests/pyplot`
- both adapted Matplotlib gallery scripts complete and emit PNGs


sselvakumaran:
after fix, grids that have another grid on its right has it's rightmost border cut off 
image
look at "not equal, looks like ellipse", "still a circle, even after changing limits" plots in the first graph and the plots in the first and second columns in the second graph. similarly ticks are cut off and aren't included in the bounding box of each plot it seems - verify pr #240 fixes this issue and show the result of applying both this and pr 240 to get a final look that is appropriate / not broken.

sselvakumaran:
Addressed the clipping report in 88d8d26.

I reproduced both exact Matplotlib 3.11 gallery examples on:

1. the previous PR #242 head; and
2. a temporary combined checkout of PR #242 + PR #240.

PR #240 did **not** repair the missing right spines/tick labels by itself. The layout-specific causes were:

- absolute PNG panels painted an opaque outer canvas, so each panel erased the right/bottom chrome of the panel before it;
- `tight_layout()` recorded options but did not reserve renderer chrome;
- `add_subplot(gs[1:, -1])` accidentally changed the figure model from 3×3 to 1×8; and
- absolute panels reserved only 8 px to the right of the plot, clipping terminal x tick labels.

The fix keeps the axes patch opaque but alpha-composites transparent outer panel chrome, preserves the original GridSpec geometry, reserves deterministic tick/title gutters for tight layout, and includes terminal ticks inside each panel canvas.

The four columns below are Matplotlib 3.11, PR #242 before, PR #242 fixed, and the fixed PR #242 with PR #240 applied:

Matplotlib 3.11 vs PR 242 before, fixed, and fixed plus PR 240

The combined column confirms the division of responsibility: this PR restores complete panel bounds/layout; PR #240 additionally improves autoscale margins and tick selection/spacing.

Validation:

- `tests/pyplot`: 542 passed, 65 skipped
- focused layout/parity tests: 27 passed
- Ruff + formatting + `git diff --check`: clean

PR #243, opened 2026-07-24
Fix pie labels and boxed text [mpl compatibility]

sselvakumaran:
## Summary

Fix the Matplotlib 3.11 pie-label and text-command gallery paths:

- preserve integer pie input dtype so `{absval:d}` formatting works
- implement `pie_label(..., rotate=True)` alignment and radial label geometry
- support named font sizes, weight, family, style, alignment, and rotation in pie text properties
- support `Axes.text(style="italic", bbox=...)` across browser, SVG, and native PNG rendering
- keep figure suptitles in positioned/native PNG figures after `subplots_adjust(...)`

This is separate from #240 and does not change autoscale margins, ticks, or bar extents.

## Gallery references

- [Pie label](https://matplotlib.org/stable/gallery/pie_and_polar_charts/pie_label.html)
- [Text commands](https://matplotlib.org/stable/gallery/text_labels_and_annotations/text_commands.html)

### Pie labels — before / Matplotlib / after

Pie label comparison

Before, the gallery aborts as soon as it requests rotated labels. After, all eight gallery figures complete. SVG keeps the radial text rotations; native PNG currently renders arbitrary-angle glyphs upright, which is a small renderer-specific text variation.

### Text commands — before / Matplotlib / after

Text commands comparison

Before, the first italic boxed label aborts the example. After, the full annotated figure renders, including the box, font styling, and figure suptitle.

## Verification

- `547 passed, 65 skipped` in `tests/pyplot`
- `pie_label.py`: all 8 PNG captures complete
- `text_commands.py`: PNG capture completes


sselvakumaran:
pie chart:
- add the rotated labels, this version adds the labels but does not rotate them.
- the pie chart is not circular (maybe another PR fixes this)
- the axes, ticks are not hidden (maybe another PR fixes this)
- pie chart has visual artifacts of thin white spokes going out of the pie, probably due to piece-wise rendering of the segments - these should not be visible and/or removed.
- the xy pie chart does not actually rotate the items (this should be fixed). the pie chart is actually circular (a different PR probably fixes this), and axes are hidden (hopefully another PR fixes this). 
text commands:
- annotations do not respect italics, mathtext italics (e=mc^2)
- box 'boxed italics text in data coords' seems to over-extend instead of cleanly wrapping the text with some padding (maybe another PR fixes this?)
- figure suptitle (subtitle?) does not respect bold (or bold at all) - also both axes title and subtitle are not centered correctly (maybe another PR fixes this?)
- arrow seems to not be large enough (maybe this is due to 2x scale?)

sselvakumaran:
I worked through each visual point in the comparison and pushed `0ca6b78`.

Exact Matplotlib 3.11 pie comparison: before XY, reference, and after #243

The comparison uses the same `pie_label` gallery capture at normalized 640×480 resolution.

What changed:

- Pie axes now use equal aspect, hide axes/spines for the default frameless pie, and cover triangle-fan antialiasing seams with same-face strokes.
- Native PNG text now supports arbitrary rotation, italic shear, and bold emphasis, so rotated pie labels, explicit italics, and the bold suptitle are preserved in static exports.
- Simple mathtext now retains italic spans only for the math variables (`E`, `mc`), across browser, SVG, and PNG.
- Text-bbox width/padding is tightened to the Matplotlib reference.
- The legacy `arrowprops={facecolor: ..., shrink: ...}` arrowhead is enlarged to match Matplotlib's YAArrow proportions.
- `startangle` was already reaching wedge geometry; it is now regression-pinned together with the corrected label rotation.

The axes title and suptitle did not need a centering change. Pixel measurement against the Matplotlib 3.11 reference puts the XY centers at 324.75/320.5 px versus 324.5/320.0 px, respectively. The apparent offset was under 0.5 px, not a 2x-DPI error.

Verification:

- 178 relevant Python/export tests passed, 3 skipped.
- 115 Rust library tests passed.
- Ruff, JS typecheck/build, Rust clippy, and diff checks passed.
- Full exact Matplotlib 3.11 scoped rerender: all 54 examples / 82 figures remain paired; automatic similarity improves from 30/82 before these fixes to 38/82 after them.

sselvakumaran:
I also checked the refreshed CodSpeed red on `0ca6b78`. It is the same cross-runtime lasso outlier appearing on several unrelated gallery PRs, not a code-path regression from this change:

- CodSpeed explicitly reports different runtime environments.
- The only reported row is `test_select_lasso_message_1m` (93.3 ms → 122.4 ms).
- The benchmark and its complete timed Python/Rust path are unchanged versus `main`: `benchmarks/test_codspeed_selection.py`, `python/xy/channel.py`, `python/xy/_figure.py`, `python/xy/interaction.py`, `python/xy/kernels.py`, and `src/kernels.rs` have no diff.
- This PR changes annotation rendering, pyplot pie/text translation, and the static raster text opcode; none are invoked inside `benchmark(channel.handle_message, uniform_figure, message)`.
- The full benchmark workflow passed; 101 other CodSpeed rows are unchanged.

So there is no causal performance fix to make in #243. The actionable rendering checks, including Matplotlib 3.11 compatibility and the full Rust/Python/JS test job, are green.

PR #244, opened 2026-07-24
Fix axis and tick-label APIs [mpl compatibility]

sselvakumaran:
## Summary

Add the missing Matplotlib 3.11 axis/tick compatibility surface used across the static gallery:

- `set_xticklabels` / `set_yticklabels`, including empty-label hiding and FixedLocator length checks
- per-axis `xaxis.grid(...)` / `yaxis.grid(...)`
- tick side and label-side visibility flags
- `rotation_mode` / `labelrotation_mode` and tick-label rotation handles
- independent x/y grid state instead of forcing the other axis off

This branch deliberately excludes all autoscale-margin behavior owned by #240.

## Gallery reference

- [Managing multiple figures in pyplot](https://matplotlib.org/stable/gallery/subplots_axes_and_figures/multiple_figs_demo.html)

### Before / Matplotlib / after

Multiple figures tick-label comparison

Before, `set_xticklabels([])` aborts the gallery and the upper axes' labels collide with the lower axes. After, both figures complete and the intended upper tick labels are hidden.

The APIs also move several additional gallery files past their prior crashes. I am not claiming those visually complete here: `boxplot_vs_violin.py` reaches an existing boxplot geometry defect, and `auto_ticks.py` still needs #240's margin API.

## Verification

- `541 passed, 65 skipped` in `tests/pyplot`
- `multiple_figs_demo.py`: both PNG captures complete
- focused axis/tick/grid tests: 34 passed


sselvakumaran:
seems like this fix is fine, but other issues that might be fixed with other PRs:
- the orange samples / line extend over the bounds in width
- tick numbers / labels are still too close to the actual ticks / axis especially on the x-axis (when text is below the axes)

sselvakumaran:
Addressed the tick-label spacing feedback in `ada03c5`.

Exact before/reference/after tick spacing comparison with 2x zoom

Matplotlib has two independent defaults here: a 3.5 pt major-tick length and a further 3.5 pt label pad. XY emitted the tick length but then placed labels with fixed renderer offsets, leaving roughly 1 px between the outward tick and its label. The browser, SVG, and PNG renderers now place labels after `outward tick length + tick padding`; pyplot supplies the rc-driven defaults and `tick_params(pad=...)` can override them, including Matplotlib-compatible negative padding.

The orange line touching the x limits is the separate autoscale-domain issue owned by #240, so I did not duplicate that change here.

Verification:

- 109 focused pyplot/style/SVG/security tests passed.
- Ruff, formatting, `git diff --check`, JS typecheck, and client build passed.
- The exact 88-example Matplotlib 3.11 scoped rerender completed with all 54 in-scope examples still paired.

I also audited the existing CodSpeed red. It only flags the unrelated `test_select_lasso_message_1m` row and explicitly warns that base/head used different runtime environments. The benchmark, selection/channel code, and native timed path are identical to `main`; this PR's tick renderer is not imported by that timed call. Local repeated 1M lasso runs stayed at a 12.7 ms median. There is no causal selection performance fix to make in this PR.

PR #245, opened 2026-07-24
Fix gallery colormaps across renderers [mpl compatibility]

sselvakumaran:
## What changed

- add Matplotlib-compatible `Reds`, `bone`, `autumn`, and `BuPu` colormaps
- keep the Python channel, pyplot, SVG/raster, and browser registries in sync
- make generic `_r` lookup work across pyplot's colormap namespace
- add cross-renderer and reversal tests so an unknown name cannot silently fall back to viridis

## Why

These are required by the in-scope Matplotlib 3.11 gallery examples. Before this change, `colorbar_basics.py` stops before drawing its `Reds_r` image because the colormap is absent.

Official reference: https://matplotlib.org/stable/gallery/color/colorbar_basics.html

Before, Matplotlib reference, and colormap-only after

The after panel is intentionally a partial execution. It proves the `Reds_r` registry/rendering fix, then stops at the next independent compatibility gap: `colorbar(location=..., anchor=..., shrink=...)`. A separate dependent PR handles those colorbar options.

## Gallery impact

- `misc/table_demo.py` now executes through its previous missing-`BuPu` blocker, but its table layout is still not visually Matplotlib-compatible, so this PR does **not** claim it as visually complete.
- `colorbar_basics.py`, `contourf_demo.py`, and `plot_streamplot.py` advance to their next independent API blockers.
- Net execution-paired delta from this isolated PR: +1. Net visually accepted delta: 0 until the downstream layout/API fixes land.

## Validation

- `570 passed, 65 skipped` across pyplot and SVG export suites
- Ruff check and format
- JavaScript typecheck and build
- exact adapted Matplotlib gallery smoke renders across native/static paths


sselvakumaran:
<NEVERMIND, FIXED IN ANOTHER PR>

second colormap (with blues and reds) is still empty.

reds looks good. the second plot with just red does not have the gradient colormap legend like the first one and should be fixed. (note the matplotlib one has the legends looking different, one is smaller than the other!)
also there is no legend for the right-most combined one, but i assume that one hasn’t been rendered at all. note the legend looks different here too.

PR #246, opened 2026-07-24
Fix colorbar gallery placement and domains [mpl compatibility]

sselvakumaran:
## What changed

- accept Matplotlib-compatible `colorbar(location=..., anchor=..., shrink=...)`
- place horizontal and vertical colorbars consistently in browser, SVG, and native PNG renderers
- render minor colorbar ticks across all three paths
- exclude masked samples from inferred colorbar domains

## Why

The official Matplotlib 3.11 `colorbar_basics.py` example uses a shrunken/anchored left colorbar and minor ticks. xy previously rejected those options; it also let masked sentinel values distort the second panel's domain.

Official reference: https://matplotlib.org/stable/gallery/color/colorbar_basics.html

Before, Matplotlib reference, and after

The final panel was rendered with this PR plus #245, because the official script also requires the independently missing `Reds_r` colormap. With both branches combined, the exact gallery script completes and the colorbar orientation, placement, domain, extension, and minor ticks are technically correct.

## Scope

This does not add `colorbar(cax=...)`; `subplots_adjust.py` remains a separate blocker. It also does not touch autoscale margins or the tick-spacing work owned by #240.

## Validation

- `544 passed, 65 skipped` across the full pyplot suite
- Ruff check and format
- TypeScript typecheck
- production JavaScript bundle build
- exact official gallery render compared with Matplotlib 3.11


sselvakumaran:
in the red plot, the the legend is smaller, but it probably should not be forced to show *more* ticks to accomodate. the red one should only have 3, bc in this version the tick spacing is too little bc theres so many ticks / numbers.

sselvakumaran:
Addressed the short red colorbar tick density in `6d53511`, with rendered-output coverage in `a6e3822`.

Correction to the earlier evidence: that image omitted #245, so it rendered the fallback blue palette and did not represent the combined gallery state. This replacement reruns the exact Matplotlib 3.11 `colorbar_basics` source in all three panels:

- **Matplotlib 3.11.0 reference**
- **Before:** #245 `2e32af3` + #246 pre-fix `64d51f6`
- **Fixed:** #245 `2e32af3` + #246 functional head `bc13c2c`

### Matplotlib 3.11 / combined before / combined fixed

Full uncropped colorbar_basics renders: Matplotlib 3.11 reference, combined #245 plus pre-fix #246 with seven crowded red ticks, and combined #245 plus fixed #246 with three readable red ticks

The renders are full and uncropped. XY is shown normalized from its native 2x export only so all three rows have the same displayed dimensions.

Verification of the corrected combined render:

- the short `Reds_r` scale has the right direction: near-zero is white at the top and approximately `-2` is dark red at the bottom
- sampled fixed endpoints are `(255, 245, 240)` at the top and `(103, 0, 13)` at the bottom
- the pre-fix SVG has seven labels: `-1.75, -1.5, -1.25, -1, -0.75, -0.5, -0.25`
- the fixed SVG has the same three major labels as Matplotlib: `-1.5, -1, -0.5`
- the image is pinned to immutable commit `650dd2b`; the asset-only cleanup head is `8e65668`, so the PNG is not present in the branch tip

The underlying #246 code at `bc13c2c` had all checks green; the two later commits only add and remove this corrected evidence image.

sselvakumaran:
Fresh CI confirms the earlier red was transient: **Test (Rust + Python + JS)** passed on `a6e3822`, including the exact previously failing **Headless render smoke (stdlib + Chromium)** step. CodSpeed also refreshed against `a6e3822` and reports **no performance alteration** (102 benchmarks untouched, 1 skipped). The Matplotlib 3.11 reference compatibility, benchmark methodology, cross-library benchmark, docs, accessibility, CodeQL, and Python 3.11 checks are green as well.

PR #247, opened 2026-07-24
Fix histogram and errorbar gallery compatibility [mpl compatibility]

sselvakumaran:
## What changed

- support histogram `facecolor`, `fill`, and scalar/per-dataset `linewidth`
- preserve Matplotlib's `step` / `stepfilled` fill override semantics
- normalize scalar histogram input as one dataset
- forward `marker`, `markersize`, `linestyle`, and dashes to the errorbar data line only
- render directional endpoint markers for `lolims`, `uplims`, `xlolims`, and `xuplims`

## Why

These gaps prevented two official Matplotlib 3.11 statistics examples from rendering at all. A script-exit-only pass initially exposed a second bug: one-sided error limits were still ordinary capped bars. This PR includes the visual geometry fix as well.

Official references:

- https://matplotlib.org/stable/gallery/statistics/errorbar_limits.html
- https://matplotlib.org/stable/gallery/statistics/histogram_normalization.html

Before, Matplotlib references, and after

## Gallery impact

- `statistics/errorbar_limits.py`: newly complete, 1/1 output
- `statistics/histogram_normalization.py`: newly complete, 9/9 outputs
- `histogram_histtypes.py` advances to the independent `rwidth` blocker
- `histogram_multihist.py` renders its first three captures, then reaches the independent `hatch` blocker

This PR does not touch the global margin/tick-spacing work in #240.

## Validation

- `544 passed, 65 skipped` across the full pyplot suite
- focused histogram/errorbar geometry and forwarding tests
- Ruff and `git diff --check`
- exact official gallery rerenders visually compared with Matplotlib 3.11


sselvakumaran:
first plot w/ error bar upper and lower limits - the CI errorbars (one vertical and two horizontals, like a capital I shape) should not be visible, just the arrows and the direct lines coming from the point should be visible. you can see this in the image provided. all five lines have this problem, but look at the blue line - the matplotlib one just has crosses with no horizontal bars besides the ones coming out of the point on the line, but the xy one also includes two more horizontals (the CI errorbars). 

also, some of the arrows do not seem correct - for the purple points the second point has an arrow going left in mol, but the xy version has the arrow’s shaft goes left but the arrowhead points right (both should point left). up and down seems fine. 

also minor nitpick but the lines seem slightly too long and not thick enough (is this a problem with 2x scale?)

in the second plot (bar plot w/ orange diamonds), the diamonds do not seem large enough (is this a problem with 2x scale?)

sselvakumaran:
Addressed the review points in `e2fb700` and `c000c4c`:

- pyplot now uses Matplotlib's default `errorbar.capsize = 0`, so the extra CI-style cap strokes are gone
- default errorbar width now comes from `lines.linewidth` (1.5 pt), and limit arrows use the normal 6 pt marker size
- left/right triangle transforms were corrected in both WebGL and native PNG rendering
- Matplotlib's `d`/`D` marker paths extend by sqrt(2) relative to `markersize`; XY had constrained them to the nominal box. PNG, SVG, and WebGL now use the Matplotlib geometry, which fixes the small orange diamonds

I reran the exact Matplotlib 3.11 gallery script. In the histogram panel, the visible diamond height at normalized 640x480 is now 6 px in both renders. The marker being half-clipped at y=0 is expected here: Matplotlib does the same because the histogram baseline is sticky. The 2x export scale was not the cause.

### Before / Matplotlib 3.11 / after

The left column is the state reviewed above; the center is the exact Matplotlib 3.11 output; the right is a fresh render from `c000c4c`.

Reviewed state, Matplotlib 3.11 reference, and current PR after

sselvakumaran:
I rechecked the current-head CodSpeed red separately from the visual review. It is not caused by this PR:

- CodSpeed explicitly flags different runtime environments.
- The only reported regression is the same unrelated `test_select_lasso_message_1m` outlier seen on other gallery PRs (93.3 ms → 124.4 ms).
- The benchmark and its complete timed Python/Rust path are byte-identical to `main`: `benchmarks/test_codspeed_selection.py`, `python/xy/channel.py`, `python/xy/_figure.py`, `python/xy/interaction.py`, `python/xy/kernels.py`, and `src/kernels.rs` have no diff.
- #247 changes histogram/errorbar/marker translation and rendering; none of those files are invoked by the lasso benchmark's timed `channel.handle_message` call.
- 101 other CodSpeed rows are unchanged and the full benchmark workflow passed.

There is therefore no causal performance fix to make in #247. The four human visual findings are covered by `e2fb700`/`c000c4c` and the exact Matplotlib 3.11 rerender evidence above.

PR #248, opened 2026-07-24
Fix line and legend gallery compatibility [mpl compatibility]

sselvakumaran:
## What changed

- support `drawstyle="steps"` as Matplotlib's `steps-pre` alias
- support 2- and 4-value `bbox_to_anchor` geometry across browser, SVG, and native PNG
- return a real `Legend`, support `Axes.add_artist()`, and render multiple explicit legend boxes
- implement `PathCollection.legend_elements()` for color and size handles
- preserve explicit scatter marker sizes in legends
- preserve scalar scatter alpha in the native affine fast path
- treat a NumPy RGBA row as one bar/legend color
- honor legend frame styles and Matplotlib's default frame alpha across renderers
- accept technically correct round dash-cap mutation

## Why

These were chained blockers in three official Matplotlib 3.11 line/marker gallery examples. Visual QA also caught two silent fidelity bugs after the scripts began exiting: native PNG dropped scatter alpha, and legend frames/styles were not actually applied.

Official references:

- https://matplotlib.org/stable/gallery/lines_bars_and_markers/horizontal_barchart_distribution.html
- https://matplotlib.org/stable/gallery/lines_bars_and_markers/scatter_with_legend.html
- https://matplotlib.org/stable/gallery/lines_bars_and_markers/step_demo.html

Before, Matplotlib references, and after

## Gallery impact

- `horizontal_barchart_distribution.py`: newly complete, 1/1 output
- `scatter_with_legend.py`: newly complete, 3/3 outputs
- `step_demo.py`: newly complete, 2/2 outputs
- `line_demo_dash_control.py` advances past round dash caps, then reaches the independent `gapcolor` blocker; `legend(handlelength=4)` is also still unsupported

The survey's remaining left tick-label clipping is the margin work owned by #240. This PR deliberately does not change global margins/tick spacing.

## Validation

- `543 passed, 65 skipped` across the full pyplot suite
- Ruff check and format
- TypeScript typecheck
- production JavaScript bundle build
- exact official gallery rerenders visually compared with Matplotlib 3.11


sselvakumaran:
for the horizontal barchart the xy one is severely cut off - only ‘estion 1’ is visible of ‘question 1’ for the y axis. as well, the borders on the top and bottom of the chart is not visible probably also due to this cutoff. the legend seems fine but it does not include the border it has (is this intentional or style? if a style choice then keep it but if it was asked for in the test then add it).

also the items in the legend seem too far apart (is this a problem with 2x scale?)

sselvakumaran:
I reproduced the issues in the exact Matplotlib 3.11 `horizontal_barchart_distribution` gallery example and pushed `c411d80`.

Before, Matplotlib 3.11 reference, and after comparison for horizontal_barchart_distribution

The three panels are saved, uncropped renders. Both XY panels intentionally retain the clipped `Question` prefix so that the separate #240 plot-layout/tick-label reservation issue remains visible rather than being hidden by this evidence.

What changed:
- legend columns now size to their own label content instead of all inheriting the longest label's width;
- handle/text/column spacing now follows Matplotlib's default proportions;
- the DOM legend gets an explicit 1 px frame (the prior PR screenshot was stale and predated the frame work);
- `xaxis.set_visible(False)` no longer suppresses the independently visible top/bottom Matplotlib spines in the browser, SVG, or PNG renderers.

The 2x export is not what caused the large spaces: those were equal-width legend cells. The new exact rerender has a visible legend frame and all four axes spines. Focused verification: 62 tests pass, Ruff passes, client typecheck/build passes, and the full 88-example gallery rerender completes with this example paired.

I also tested the published #240 head (`3b3c2ab`) merged with this head. The combined render still clips every y category label to `estion 1`, etc.; #240 currently changes data autoscale margins but does not reserve layout width for categorical tick-label text. I have kept that separate from this PR rather than duplicating its plot-padding work.

sselvakumaran:
Post-`c411d80` audit:

- There are no submitted reviews, inline review threads, or newer human comments after the existing rendering evidence.
- I rechecked the exact official `horizontal_barchart_distribution` output against the rerender whose manifest records `Matplotlib 3.11.0`. The legend frame, per-column spacing, and all four spines now match the requested behavior. The remaining clipped `Question` prefix is the separate plot-layout/tick-label reservation issue already called out for #240; it is not a remaining line/legend regression.
- The CodSpeed lasso result is not attributable to this PR. The benchmark and its complete timed path are byte-identical between `main` and `c411d80`: `benchmarks/test_codspeed_selection.py`, `python/xy/channel.py`, `python/xy/_figure.py`, `python/xy/interaction.py`, `python/xy/kernels.py`, and `src/kernels.rs` all have identical Git object IDs. This PR changes legend/renderer code only, the other 101 benchmarks are unchanged, and CodSpeed explicitly reports that base and head ran in different runtime environments. I therefore did not add an unrelated selection-path change.
- The initially canceled queued jobs were infrastructure rather than test failures. I reran them: [Install without Rust](https://github.com/reflex-dev/xy/actions/runs/30108572653/job/89536603500), [Docs tests and lint](https://github.com/reflex-dev/xy/actions/runs/30108572630/job/89536607965), and production docs on [Python 3.11](https://github.com/reflex-dev/xy/actions/runs/30108572630/job/89536888037) and [Python 3.12](https://github.com/reflex-dev/xy/actions/runs/30108572630/job/89536888196) all pass. The two canceled dynamic CodeQL shards never started and GitHub does not permit retrying that dynamic run; the Python and JavaScript/TypeScript CodeQL shards passed and the aggregate CodeQL check is neutral.

No additional non-#240 code fix was warranted, so the PR remains clean at `c411d80`.

PR #252, opened 2026-07-24
Fix step histogram autoscaling [mpl compatibility]

sselvakumaran:
## Summary

- teach the pyplot autoscaler that compact `stairs(values, edges)` entries contribute bin edges on x and histogram heights on y
- include filled-area geometry and its baseline in the same autoscale/scale-transform path
- add regressions for step and stepfilled histograms, density normalization, and probability weights

## Root cause

`Axes.hist(..., histtype="step")` correctly computed the bin counts, density, and weights, but stored the result as the compact `stairs(values, edges)` mark. Pyplot's entry scanner did not understand that reversed argument order and classified stairs-only panels as dataless, pinning both axes to the empty `(0, 1)` range. This clipped the real `[-4, 4]` distributions in Matplotlib gallery `statistics/histogram_normalization.py` figures 7–8.

## Validation

- `550 passed, 65 skipped` across `tests/pyplot`
- Ruff check and format clean
- exact Matplotlib 3.11 88-example gallery harness: 65 examples / 105 figures paired at the current union head, with figures 7–8 restored and no gallery-wide count regression

A Matplotlib / before / after comparison is attached in the PR comments.


sselvakumaran:
Exact gallery verification against Matplotlib 3.11:

Matplotlib 3.11 vs current union vs fixed histogram normalization

The middle column is the authoritative current-head union of #206, #240–#248, and #211. The right column applies this PR on top of that union.

The histogram counts were already numerically correct; the compact `stairs(values, edges)` entry was omitted from pyplot autoscaling, so stairs-only panels were pinned to the empty `(0, 1)` view. The fix restores the real `[-4, 4]` bin domain. Regression tests also prove `sum(density * bin_width) == 1` and probability weights sum to one.

The remaining inter-panel ylabel spacing visible in output 008 is independent grid-layout work, so I am keeping it out of this technical autoscale PR and treating it as a separate layout fix.

PR #253, opened 2026-07-24
Fix categorical pyplot gallery rendering [mpl compatibility]

sselvakumaran:
## Summary

- map pyplot categorical coordinates to first-seen ordinal positions while computing view limits, matching the core's category mapping
- support Matplotlib's per-patch `bar(..., label=[...])` legend semantics, including underscore filtering and each patch's resolved color
- keep bar geometry vectorized and add focused gallery regression coverage for categorical trajectories, legends, and `barh` widths

## Root causes

Pyplot's pre-build autoscale scan discarded string coordinates, so categorical line and scatter plots fell back to a dataless `(0, 1)` view and clipped every category after the second. Separately, a sequence passed as a bar label was stringified as one container label instead of being attached to the individual bar patches.

## Visual comparison

Columns are Matplotlib 3.11, the branch base, and this PR. The first row shows the per-bar legend fix; the second shows all six categorical trajectories restored.

Matplotlib, before, and after categorical gallery comparison

## Validation

- `PYTHONPATH=python /Users/srimans/dev/reviz/.venv/bin/python -m pytest -q tests/pyplot/test_categorical_gallery_regressions.py` — 5 passed
- `PYTHONPATH=python /Users/srimans/dev/reviz/.venv/bin/python -m pytest -q tests/pyplot` — 558 passed, 65 skipped
- Ruff check and format check pass
- exact Matplotlib 3.11 gallery sources rendered for `bar_colors.py` and `categorical_variables.py`


sselvakumaran:
maybe a problem solved by another PR - labels should ideally go where there is no lines if possible; like the label here should go in the upper left or lower right of the graph. otherwise this specific fix works.

PR #261, opened 2026-07-24
Fail browser probes when chromium crashes, instead of skipping

Alek99:
`run_browser_probe` could not distinguish *"chromium is missing"* from *"chromium ran and died"*. `_dump_dom` returned `None` for both, so three crashed attempts left `last` unset and fell through to `pytest.skip`.

The effect: a client regression that hard-crashes the renderer — a bad shader, a null deref in `renderStandalone`, a broken bundle — turned **all 36 browser tests green**. Under `make test` those skips print as a bare `s` and the run exits 0. The worse the breakage, the more reliably it was hidden.

**Confirmed, not inferred:** pointing `XY_CHROMIUM` at a script that `exit 1`s gave `1 skipped`, exit 0.

## Change

`_dump_dom` now returns `(dom, failure)` and separates the two cases:

| Situation | Before | After |
|---|---|---|
| Crash / non-zero exit | skip | **fail** |
| Timeout | skip | **fail** |
| Probe error, missing result attribute | fail | fail |
| Cannot spawn the binary at all | skip | skip, with the exec error as the reason |
| Cannot spawn, `XY_REQUIRE_BROWSER=1` | skip | **fail** |

The trailing unconditional `pytest.skip` is gone — every way of not getting a result now fails. CI exports `XY_REQUIRE_BROWSER=1` beside the existing `XY_CHROMIUM` so the browser layer cannot pass by absence.

## Verification

- Crashing browser now **fails** (was: `1 skipped`, exit 0).
- Unlaunchable browser skips with the exec error as the reason, and **fails** under `XY_REQUIRE_BROWSER=1`.
- Full suite on this base: **2222 passed, 4 skipped, 0 failed** — unchanged.
- `ruff check`, `ruff format --check`, and `scripts/verify_ci_workflow.py` pass.

## Not covered here

22 tests skip at their own `find_chromium() is None` check *before* reaching `conftest`, so `XY_REQUIRE_BROWSER` does not fire for the "no browser installed at all" case — which is the `python_floor` job's situation, since it sets no `XY_CHROMIUM` and installs no Playwright. Rewiring those call sites is separate work.

## Summary by CodeRabbit

* **Tests**
  * Improved browser-based test reliability by distinguishing unavailable browsers from temporary launch or rendering failures.
  * Browser test issues now provide clearer failure messages and retry transient failures before reporting an error.
  * Environments without a usable browser are handled explicitly, with stricter validation in continuous integration.

* **Chores**
  * Updated automated checks to ensure browser-dependent functionality is validated consistently.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -167,6 +167,10 @@ jobs:
         # full Figure.to_png path runs here rather than skipping.
         run: |
           export XY_CHROMIUM="$(node -e "console.log(require('playwright').chromium.executablePath())")"
+          # Chromium is installed here, so a probe that cannot launch it is a
+          # defect: fail rather than skip, so the browser layer cannot pass by
+          # absence. (A crash or non-zero exit already fails unconditionally.)
+          export XY_REQUIRE_BROWSER=1
           .venv/bin/pytest -q
 
       # End-to-end: a real Figure (line decimation + scatter) through to_html's
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 
 import html
 import json
+import os
 import re
 import subprocess
 from pathlib import Path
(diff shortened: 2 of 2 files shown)


Alek99:
Superseded by #262 — GitHub auto-closed this when the branch was renamed to `fix/browser-probe-crash-fails-not-skips` to match the narrowed scope. Same commit, same diff.

PR #265, opened 2026-07-24
Cast lasso rays in the core instead of once per polygon edge

Alek99:
## What

`select_polygon` bounded the work correctly — zone-pruned bbox first, ray casting only over those candidates — and then spent all of it in NumPy, one vectorized pass per polygon **edge**. Each pass materialized full-length crossing and edge-intersection temporaries over every candidate, so a 64-vertex lasso over 160k candidates moved roughly half a gigabyte through memory to answer a question that fits in registers. Worse, cost grew linearly with vertex count: a 2048-vertex freehand lasso — the documented API ceiling — took ~370 ms. Every other selection primitive was already native; this one was not.

`kernels.polygon_select` walks edges inside the point loop. Crossing parity is order-independent, so the answer is unchanged.

On top of that it buckets edges into y slabs. An edge can only flip parity for a point inside its own y-extent, so the index hands back a *superset* of the edges that can cross a given row and the rest are provably no-ops — restricting the candidate set is not an approximation. That is what flattens the vertex scaling.

## Numbers

1M uniform points, 160k bbox candidates, min-of-3:

| vertices | numpy | native | speedup |
|---:|---:|---:|---:|
| 16 | 3.70 ms | 2.56 ms | 1.4x |
| 64 | 12.40 ms | 3.20 ms | 3.9x |
| 256 | 48.30 ms | 3.03 ms | 15.9x |
| 1024 | 188.18 ms | 4.47 ms | 42.1x |
| 2048 | 377.57 ms | 5.39 ms | 70.0x |

The full gesture unit `test_select_lasso_message_1m` measures — dispatch, bbox prune, cast, wire mask reply — goes **13.59 ms -> 3.21 ms** (min-of-5).

There is a second reason to want this. `test_select_lasso_message_1m` has been the suite's largest source of false alarms: it is bimodal at ~90 ms and ~123-141 ms and flips on changes that cannot reach it — docs-only #251 read "+39%", CI-only #236 read "-25%", and #264 (which touches only `marks.py` and `_figure.py`) just read "-27%", each with the bot's "different runtime environments" warning attached. A memory-bound NumPy loop is exactly the workload whose modeled cost swings when the runner changes; collapsing it to a register-resident scan should shrink that sensitivity considerably.

## Correctness

The arithmetic is preserved term for term. `x_j - x_i` and `y_j - y_i` are hoisted per edge — exact subtractions — but the division is **not** strength-reduced to a reciprocal multiply, which would not be. The guarded division covers exactly the lanes the vectorized form computed unconditionally and then masked its inf/NaN away, since endpoints straddling the ray cannot have equal y. Bucketing is declined for a non-finite polygon, where slab bounds stop meaning anything, and below 16 vertices, where it would not pay for itself.

Verified identical to the previous predicate on:

- convex, concave (star), and self-intersecting (bowtie — even-odd parity is load-bearing) polygons
- axis-aligned boxes, whose horizontal edges are the masked-division lanes
- collinear, zero-height, degenerate and repeated-vertex polygons
- points landing exactly on vertices and edges
- NaN and +-inf coordinates
- sparse candidate subsets (the real call shape), with canonical ascending order preserved
- both sides of the 16-vertex index gate, and up to the 2048-vertex ceiling

Both the Rust unit test and `tests/test_select_polygon.py` carry the per-edge formulation as an executable reference rather than describing it in a comment.

- `cargo test` — 117 passed; `cargo clippy` clean; no new `cargo fmt` diffs
- full Python suite — 2240 passed, 4 skipped
- `scripts/abi_smoke.py` — 133 checks passed
- ABI 40 -> 41, bumped in `src/lib.rs` and `python/xy/_native.py` together
- placement audit in `spec/design/rust-engine.md` updated

## Summary by CodeRabbit

* **New Features**
  * Added native-accelerated “lasso” polygon selection for points, using candidate row subsets and returning matching point IDs in order.

* **Bug Fixes**
  * Improved selection consistency across concave, self-intersecting, degenerate, and non-finite geometries.

* **Tests**
  * Added extensive pytest coverage for native polygon selection accuracy, subset behavior, ordering, and input validation (including invalid polygon/row/array shapes and bounds).


Alek99:
Both review findings were real and are fixed in 6401310.

**Zero-length `from_raw_parts` (`src/lib.rs`).** Correct — the polygon slices were built before the vertex count was checked, so a caller passing null for an empty polygon reached `slice::from_raw_parts(null, 0)`, which requires a non-null pointer even at length zero. The "fewer than three vertices encloses nothing" case is now answered before any slice exists.

**Unbounded CSR (`src/kernels.rs`).** Also correct, and the reasoning about `select_polygon`'s 2048-vertex cap not covering the kernel's own contract is right.

I took a different fix than the suggested `16..=1<<16` ceiling, though. Refusing to index above a vertex ceiling sends exactly the largest polygons — the ones that hurt most — to the fallback scan, which is O(edges) for *every* point; at 100k vertices over 160k candidates that is worse than what this PR replaced. Instead the CSR is capped at 2²⁰ entries and the slab count is shed to stay inside it, so a huge polygon keeps an index with coarser buckets. Below 8 slabs the index no longer earns its build and declines. That bounds the allocation at 4 MB and keeps the u32 cursors well clear of overflow, while preserving the property the PR is about.

Neither change can move an answer: the budget only coarsens bucketing, and bucketing only ever restricts the tested edges to a superset of those that can cross the row.

Added coverage: a Rust test pinning the CSR invariant at 16 / 2048 / 65536 / 200000 vertices with every edge spanning the full height (the worst case for the index), and `ring_2048` — the API ceiling — added to the Python equivalence sweep.

On the spec nitpick: thresholds are now recorded in `spec/design/rust-engine.md` (the < 16 gate, the 256-slab cap, the record budget, and the non-finite fallback).

Re-verified: 118 Rust tests, 2241 Python tests, clippy clean, no new `cargo fmt` diffs, ABI smoke 133 checks. The 2048-vertex speedup is unchanged at 74x.

PR #276, opened 2026-07-25
Deploy docs on semver release tags instead of CalVer

Kastier1:
`deploy-docs-stg.yml` was the only workflow still keyed to the flexgen/helm-charts CalVer scheme (`2026.WW.N`). Every other workflow here runs on push/PR/dispatch, and `release.yml` runs on the library's `v*` tags. This aligns the docs deploy with that.

## Changes

- **Trigger:** `tags: ["20[0-9][0-9].[0-9]*"]` → `tags: ["v*"]` — the same tags `release.yml` publishes to PyPI from, so the docs site ships the version it documents.
- **Version guard:** `^20[0-9]{2}\.[0-9]+\.[0-9]+$` → `^v[0-9]+\.[0-9]+\.[0-9]+[A-Za-z0-9.-]*$`. Accepts `v0.1.0`, `v1.20.3rc1`, `v0.1.0-beta.1`; rejects `v1.2` and PEP 440 local versions (`v1.2.3+local`) — `+` is not a legal Docker tag character and this value becomes the image tag.
- **`workflow_dispatch` input description** updated to match.
- **Removed the `Create GitHub Release` job**, rewiring `helm-pr-prod` off it.

## Why the release job had to go

On a `v*` tag, `release.yml`'s `publish-pyodide` already does `gh release create "$tag" --generate-notes`, gated on tag == pyproject == CHANGELOG. Leaving this workflow's release job in place meant two workflows racing to create the same release. The race is unlikely but not impossible — the docs job runs after prod approval, the wheel matrix is slow — and if the docs deploy won, the library release would permanently carry notes reading "Production deployment of XY docs" instead of generated changelog notes.

## Behavior change to weigh

Docs prod deploys are now coupled to library releases: cutting a CalVer tag no longer ships a docs-only prod update. The remaining paths are `workflow_dispatch` against an existing `v*` tag, and `deploy-docs-dev.yml`, which still tracks `main` on every push. If docs should keep shipping independently of PyPI releases, the fix is to keep both trigger patterns instead — say the word and I'll add the CalVer pattern back alongside `v*`.

## Verification

- `make check-ci` passes.
- Workflow YAML parses; job graph is `prepare → build → helm-pr-stg → await-prod-approval → helm-pr-prod` with no dangling `needs`.
- Version regex exercised against `v0.1.0`, `v0.0.2`, `v1.20.3rc1`, `v0.1.0-beta.1` (pass) and `2026.26.1`, `v1.2`, `v1.2.3+local`, `v1.2.3;rm` (reject).
- `pre-commit run --all-files`, `ruff check .`, `ruff format --check .` all clean.

Note: no `spec/` or `docs/` page documents the deploy tag scheme, so there was nothing to sync there.

## Summary by CodeRabbit

* **Chores**
  * Updated documentation staging and production deployment to trigger only on semantic version `v*` release tags.
  * Enhanced version validation to reject unsupported tag formats and overly long versions.
  * Removed automatic GitHub Release creation from the docs deployment flow.
  * Improved production release safety by verifying the corresponding GitHub Release and package availability before promoting Helm/production updates.

### DIFF
--- a/.github/workflows/deploy-docs-stg.yml
+++ b/.github/workflows/deploy-docs-stg.yml
@@ -7,15 +7,15 @@ on:
   workflow_dispatch:
     inputs:
       version:
-        description: "Existing XY docs deploy tag (CalVer, for example 2026.26.1)"
+        description: "Existing XY release tag (semver, for example v0.1.0)"
         required: true
         type: string
   push:
-    # CalVer deploy tags (2026.WW.N) — matches the flexgen/helm-charts release
-    # scheme and stays disjoint from the library's semver v* tags, which
-    # trigger release.yml (PyPI/Cargo) and must NOT deploy the docs site.
-    tags:
-      - "20[0-9][0-9].[0-9]*"
+    # The library's semver release tags — the same `v*` tags release.yml
+    # publishes to PyPI from, so the docs site ships the version it documents.
+    # Deploying an arbitrary commit outside a release is workflow_dispatch's
+    # job (or deploy-docs-dev.yml, which tracks main on every push).
+    tags: ["v*"]
 
 permissions:
   contents: read
@@ -42,8 +42,11 @@ jobs:
           EVENT_SHA: ${{ github.sha }}
           REPO: ${{ github.repository }}
         run: |
-          if [[ ! "$VERSION" =~ ^20[0-9]{2}\.[0-9]+\.[0-9]+$ ]]; then
-            echo "::error::version must be CalVer like 2026.26.1"
+          # Same shape release.yml gates on (tag == v<pyproject version>),
+          # minus PEP 440 local versions: `+` is not a legal Docker tag
+          # character and this value becomes the image tag.
+          if [[ ! "$VERSION" =~ ^v[0-9]+\.[0-9]+\.[0-9]+[A-Za-z0-9.-]*$ ]]; then
+            echo "::error::version must be a semver release tag like v0.1.0"
             exit 1
           fi
           if [[ "$EVENT_NAME" == "workflow_dispatch" ]]; then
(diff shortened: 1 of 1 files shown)


Kastier1:
Went through the bot review comments and traced the deploy chain against `reflex-dev/helm-charts` and `reflex-dev/infra`. Pushed a6d8b94 with the follow-ups.

## Review comments

**CodeRabbit — "Enforce actual SemVer and Docker tag limits" — partially applied.**

The length check is right and is now enforced up front (128 chars, matching what `_build-docs-images.yml` and `_helm-docs-pr.yml` already check downstream), and the pattern no longer permits a trailing separator (`v1.2.3-`).

The strict-SemVer part I did not take, deliberately. This repo's release contract is `scripts/check_release_version.py`, which requires `tag == "v" + <pyproject version>` — a PEP 440 version, not SemVer. An rc here tags as `v0.2.0rc1` (no dash) and a post-release as `v0.2.0.post1`; the proposed pattern rejects both, and `v1.2.3.4` is a legal PEP 440 version too. Adopting it would block valid release tags from deploying. Also flagging that the committable suggestion is malformed — it renders as `${`#VERSION`}`, with backticks inside the parameter expansion, which is a bash syntax error if applied as-is.

**Codex — "Enforce the repository's release-version gate" — applied.** Correct: shape is not releasability. `prepare` now checks out the tagged tree and runs the same `scripts/check_release_version.py` gate `release.yml` runs, so a stray `v9.9.9` can't deploy docs for a version that can never be published.

**Codex — "Wait for the library release before promoting production" — applied.** This is the flip side of dropping the release-creation job, and I'd missed it. New `verify-library-release` job sits between prod approval and `helm-pr-prod`: it polls until the GitHub Release exists *and* the version resolves on PyPI (30 × 60s), and fails otherwise. Checking PyPI as well as the release matters because `publish-pyodide` creates the GitHub Release but `publish` is what puts the wheel on PyPI — they can diverge.

## Helm / infra interaction

Traced end to end: xy tag → `_build-docs-images.yml` (ECR `xy/frontend`,`xy/backend`) → `_helm-docs-pr.yml` (`charts/internal/values-{stg,prod}.yaml`) → helm-charts `release-and-pin.yml` → infra `k8s/clusters/*/00-sources.yaml` → Flux.

**Nothing breaks.** Specifics:

- **The chart never parses the tag.** `charts/internal/templates/_helpers.tpl:94` only interpolates it into an image reference; it isn't used as a Kubernetes label or version. `v0.1.0` behaves exactly as `2026.29.1` did.
- **`yq ... strenv(IMAGE_TAG)` is safe**, and marginally safer than before — a `v`-prefixed tag can never be YAML-coerced to a number, whereas a two-component CalVer tag could.
- **The helm→infra chain is title-driven, not tag-driven.** `release-and-pin.yml` fires on merged `helm-deploy-bot[bot]` PRs whose title contains `stg]`/`prod]`; our title is `[internal/xy stg] Update image to <tag>`, so it still matches.
- **No collision with the infra pin grep.** `release-and-pin.yml` does `grep -oE 'tag: v[0-9.]+' | head -1` against `infra/k8s/clusters/*/00-sources.yaml`. Those files contain only the helm-charts GitRepository pin (currently `tag: v2026.30.4`) — xy image tags live in helm-charts values files, which that grep never reads. Worth stating explicitly since both are now `v`-prefixed. Infra references xy in exactly one place, a Vault secret sync, with no image tags.

**Two things worth a decision, neither blocking:**

1. **xy becomes the only non-CalVer tag in the internal chart.** Every neighbour in `values-stg.yaml` / `values-prod.yaml` is CalVer (`2026.29.0`, `2026.28.1`, …), and `blog` uses `r-2026-07-01-1`. Cosmetic, but the original workflow comment cited fleet consistency as the reason for CalVer, so it is a real (small) cost being paid.

2. **Docs can no longer ship without a library release** — the point from the PR description, now sharper given `verify-library-release`. A docs-only fix on an already-released version has no path to prod: re-running `workflow_dispatch` on `v0.1.0` rebuilds from the same tag, and if the ECR repos are immutable the push fails outright (I could not verify mutability — the ECR account is AWS and `infra` is the AKS cluster, so the repo config isn't in either checkout). If docs need to ship independently, the fix is to keep the CalVer pattern alongside `v*` rather than replacing it. Happy to do that instead.

masenf (approved):

PR #302, opened 2026-07-26
1/7 Widen the mark CSS subset: stroke-linecap and marker-shape

Alek99:
First of a seven-PR stack that makes claims about xy's customizability defensible instead of assertable. This one ships capability; the later ones make it measurable and findable.

## What changes

**`marker-shape`** was nearly free. All three renderers have drawn 17 marker symbols for some time and `symbol=` already reached every one — only the CSS spelling was missing. Both spellings compile to the same trace-style value, so they build byte-identical specs.

**`stroke-linecap`** was not free, and the reason is the interesting part: the three renderers disagreed and none of them said so.

| | before | after |
|---|---|---|
| native rasterizer | round (clamped segment distance field is a capsule) | selectable, defaults round |
| WebGL client | butt, with a half-pixel bleed | selectable, defaults round |
| SVG writer | `round` hardcoded on line paths; area outline inherited SVG's `butt` | explicit on every stroked path |
| PDF | read `butt` back out of the SVG | follows the SVG |

`round` is now the contract everywhere, chosen because the rasterizer is the reference for static export.

The rasterizer keeps its existing fast path untouched for round and routes anything else through `stroke_shaped`, which decomposes a polyline into oriented-box segments plus join primitives (disc, bevel triangle, miter wedge with SVG's limit of 4), max-combined into one scratch buffer so overlapping join geometry cannot double-darken a translucent stroke. Join geometry is not optional there — butt-capped segments leave a notch at every interior vertex.

## What is deliberately not here

**`stroke-linejoin` does not ship.** SVG emits it and the rasterizer implements all three joins, but the WebGL client draws a polyline as one instanced quad per segment with no join geometry at all. An attempt at a second instanced pass over interior vertices did not survive verification (its vertex attributes read as NaN; bisected to the attribute fetch, not the shader math), so shipping the property would mean two renderers honoring it and one ignoring it — precisely the failure `styles.py` exists to prevent.

The default-join divergence that leaves behind — rasterizer fills the vertex, WebGL leaves the notch — predates this change and is now written down instead of left to be discovered. PR 4 records it as a registry row.

## Verification

Caps are checked end to end, not just at the seam:
- a Rust unit test on coverage past the endpoint, and that `round` is unchanged
- a Python test on rasterized ink (`square` paints more than `butt`)
- three Chromium screenshots that hash differently per cap

Absent keys stay off the wire, so specs built before this change are byte-identical.

`cargo test` 119 passed · `uv run pytest` 2326 passed · `node js/build.mjs` clean · `render_smoke_nonumpy.py` OK

## Note for reviewers

`ABI_VERSION` 41 → 42: `OP_STROKE` and `OP_SMOOTH_STROKE` carry two more bytes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01N8awRC7yTDToVG7jMvjmcx)_

## Summary by CodeRabbit

- **New Features**
  - Added `stroke-linecap` styling for line-based marks, supporting butt, round, and square caps.
  - Added `marker-shape` styling for scatter marks as an alternative to symbol selection.

- **Bug Fixes**
  - Standardized line-cap rendering across raster, WebGL, and SVG outputs, with a consistent round default.
  - Improved cap and dash rendering for stroked and dashed lines.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,17 @@ in the README).
 ## [Unreleased]
 
 ### Added
+- Mark `style=` accepts **`stroke-linecap`** (`butt`/`round`/`square`) on the
+  line family, with the standard SVG semantics: it shapes the polyline's two
+  ends and each dash end. Only the line family takes it, because it describes
+  stroked open-path geometry. XY's default is `round`, not CSS's `butt`, so
+  existing specs are byte-identical. `stroke-linejoin` is deliberately not
+  offered yet — the SVG and native-raster writers implement it but the WebGL
+  client has no join geometry, and a property two renderers honor and one
+  ignores is exactly what the mark style subset exists to prevent.
+- Mark `style=` accepts **`marker-shape`** on `scatter`, the CSS spelling of the
+  existing `symbol=` argument. Both resolve to the same trace-style value, so
+  the two spellings build identical specs.
 - `colormap=` accepts a **custom ramp** built from your own colors, not only one
   of the twenty built-in names: a sequence of 2–256 CSS colors, `(position,
   color)` pairs, or a CSS `linear-gradient(...)`. Every form resolves once, in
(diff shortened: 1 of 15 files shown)


Alek99:
Closing in favour of #310, which carries this same work as a single PR with before/after renders of the cap fix.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #309, opened 2026-07-26
Cut peak memory across the encode and payload paths

Alek99:
Audit of engine memory, plus the wins from it that survived a perf gate.

Instrumented peak RSS and `tracemalloc` per workload in fresh processes. The finding that
shaped the PR: **the payload path is already lean** — a 10M-point density build costs
13.8 MB over the 160 MB of canonical data, and a 10M-point M4 line costs 0.9 MB. Nothing
there needed fixing. Every real problem was a *transient*, and almost all of them were in
the **encoders**. Since freed memory is not returned to the OS, each one-shot peak
permanently raises the process high-water, so cutting a peak cuts steady-state RSS too.

Three commits, kept unsquashed — they are independently validated and independently
revertable.

## Memory: before → after

| path | before | after | |
|---|---|---|---|
| JPEG encode, photographic 1800×840 | 1515.9 MB | **399.9 MB** | −74% |
| JPEG encode, gradient 1800×840 | 127.2 MB | **48.4 MB** | −62% |
| JPEG encode, chart 1800×840 | 108.2 MB | **48.4 MB** | −55% |
| PNG encode, indexed 3840×2160 | 273.7 MB | **33.2 MB** | −88% |
| PNG encode, indexed 1800×840 | 49.9 MB | **6.1 MB** | −88% |
| PNG encode, truecolor 1800×840 | 36.3 MB | 30.3 MB | −17% |
| 2.1M-row continuous color channel resolve | 44.1 MB | **7.3 MB** | −83% |
| colored 2.1M-point first paint (RSS) | 181.4 MB | **145.8 MB** | −19% |
| `to_svg`, 100k-point scatter | 38.7 MB (100.1 MB RSS) | 26.9 MB (84.8 MB RSS) | −30% |
| `to_html`, 1M-point scatter | 41.3 MB | 34.0 MB | −18% |
| streamed figure's *reported* RAM | 1.92 of a real 3.20 MB | exact | 40% was hidden |

## Speed: the same changes, measured in isolation (min-of-N, interleaved)

| path | before | after |
|---|---|---|
| PNG encode, indexed 3840×2160 | 86.9 ms | **40.2 ms** |
| PNG encode, indexed 1800×840 | 15.3 ms | **8.6 ms** |
| direct-tier scatter payload emit, 100k | 0.058 ms | **0.038 ms** |
| JPEG encode, gradient 1800×840 | 51.0 ms | **41.5 ms** |
| JPEG encode, chart 1800×840 | 40.3 ms | 38.3 ms |
| `to_html`, 1M | 8.36 ms | 8.00 ms |

## What was actually wrong

Every one of these was "copying what the caller already holds":

- **The JPEG entropy packer built two int64 index vectors per output *bit*** — 17 bytes per
  bit — so a 2.8 MB stream peaked over 1.5 GB. It now packs in bounded bit passes, carrying
  each pass's sub-byte remainder so the emitted bytes are exactly those of the whole-image
  pass (0xFF stuffing is byte-local, so it stays per-pass). Around it, planes are released
  as consumed, the quantize chain runs in place, and token fields are freed as they are
  gathered into MCU order. The rounding rule is now `trunc(x + copysign(0.5, x))`: same
  round-half-away-from-zero in three in-place passes instead of four passes and four
  full-size temporaries, and exact rather than approximate — IEEE addition is
  sign-symmetric, so for `x < 0` the sum is the exact negation of `|x| + 0.5`.
- **The indexed-palette PNG encoder held five full frames**: an `intp`-per-pixel
  `np.unique` inverse (twice the image itself), a redundant `astype`, a list of per-row
  `bytes`, its join, and a `tobytes`. Now: one `(h, stride+1)` staging buffer,
  `searchsorted` over the ≤256-entry palette in row blocks, buffer handed straight to zlib.
- **Color quantization was nominally chunk-bounded at 2²² = 4M rows**, so every real column
  still ran as one shot with three simultaneous full-length f64 temporaries. Chunked at 2¹⁸
  with in-place stages; `direct_rgba` shipping (previously unchunked) routed through the
  same helper.
- **Direct-tier scatter/line/area rebuilt an all-visible row mask** on every build — two
  `isfinite` passes, an `&`, an `np.all`, two N-byte temporaries. Zone maps count NaN *and*
  ±inf as null (§22/§19), so with linear axes and no nulls the mask cannot drop a row. Area
  additionally built an identity index vector and gathered animation keys through it.
- **SVG joined the mark markup twice** (once inside the return f-string) and kept one Python
  `str` per point, ~50 bytes of object header each. One flat join; markers collapse into
  4096-marker blocks.
- **HTML export copied the base64 payload twice more than needed.** Chunks are appended as
  their own document parts and joined once. The newline stays *between* blocks, so a
  payload-free document keeps its exact byte layout.
- **Small ones**: the rasterizer's serial point/segment passes no longer collect `0..n` into
  a `Vec<u32>` just to iterate it in order (the band painters take an item iterator, so
  fanned-out callers pass bucketed indices and serial callers pass a range);
  `render_raster` borrows the display list instead of freezing a `bytes` copy of it; facet
  PNG export drops two full-frame `tobytes()`; the density quantizer keeps one temporary.

## Report accuracy (§27)

`memory_report()` now reports `capacity_bytes` per column and `canonical_capacity_bytes` per
store, and builds `resident_array_bytes` from the capacity total. A streamed column's
`values` is a prefix view of its capacity-doubling growth buffer, so up to half of what the
process holds was invisible to the report — a 20-append figure reported 1.92 MB of a real
3.20 MB. Continuous channels already reported their growth buffers this way; columns now
match. Figures that never appended report exactly what they did before. Dossier §27 updated.

## Validation

**Output is byte-identical everywhere.** Fingerprints hashed on `origin/main` and on this
branch and diffed:

- 89 payload / split-payload / density-view / drill-cycle / selection / append-frame /
  SVG / PNG / PDF / JPEG / WebP / HTML fingerprints across 30 chart configurations,
  including NaN and ±inf rows, log and symlog axes, wide categorical, and direct-RGBA
  channels
- 90 JPEG encodes (5 qualities × 18 images: flat/noise/gradient at five sizes down to 1×1,
  non-multiple-of-8 dimensions, RGB-vs-RGBA input, all-black, all-white)
- 30 HTML documents, including animation-progress and custom-CSS variants and an empty figure
- animation/transition-key paths for scatter, line and area, with and without non-finite rows

The only intentional difference is `memory_report`, which gained fields.

`pytest -q`: 2392 passed, 4 skipped. `cargo test`: 118 passed. clippy, `ruff check`,
`ruff format --check` clean; `ty check` reports the same 13 pre-existing diagnostics as the
base commit.

**Perf gate**: three interleaved A/B campaigns over the 89-benchmark CodSpeed suite in
walltime mode — arms alternated (`ref mem mem ref ref mem`), full suite per arm,
min-of-runs per arm, 16 suite runs total, and the native dylib swapped per arm alongside the
Python tree. Final campaign: **median +0.24%, mean −0.49%, 0 of 89 rows slower by >5%.**

## Note for the CodSpeed comment

Expect noise on rows this branch cannot touch. Every single-campaign outlier above +5%
failed to reproduce:

- `contour_core_2d` read +9.3% in one campaign and +2.1% in another; it has ~17% intra-arm
  spread, and in an isolated in-process interleaved benchmark this branch produced the
  fastest run of all six (3.43 ms vs the baseline's best 4.65 ms).
- `weighted_ecdf_native` (+6.4% in one campaign) and `sample_mask` (+4.7%) are untouched
  native kernels running from the *same* dylib in both arms — noise by construction.
- `native_png_export_scatter` read +0.79% in-suite while an in-process interleaved
  micro-benchmark showed equal-or-better medians and the segments path reproducibly −0.8%.
  Reporting it rather than hiding it.
- `select_lasso_message_1m` is the repo's known bimodal row.

## Deliberately not in this PR

Documented rather than coded, because each is a trade or a bigger change than a memory fix
should smuggle in:

- **The colored pyramid accumulator is the largest single number in the engine.**
  `MeanColorCell` is 40 B/cell (`kernels.rs`), and `tiles::build_color` holds it alive while
  materializing the count and color planes: a 168 MB transient at the 2048² default for a
  2.1M-point trace, and `MEAN_COLOR_ACCUM_BUDGET_BYTES` (1 GB) permits 4 workers — 672 MB on
  a row-heavy trace. The exact u64 sums are a documented bitwise-determinism guarantee, so
  the cell cannot shrink without changing that contract. Both available levers (lower the
  worker budget; pick `base_dim` from row count) regress real-world build wall-clock that no
  benchmark covers, so they are trades, not wins.
- **`range_indices`/`bin_2d_indices` allocate 4N regardless of result size** — a 400 MB
  transient at 100M rows to return a few KB for a deep-zoom window. Chunking adds two extra
  copies of the result exactly when the result is the whole column; count-then-fill doubles
  a memory-bound scan. The sound fix is structural: `interaction.py` scans the view window
  and `_padded_drill_window` then scans up to three *superset* windows, discarding all but
  the winner — one scan of the widest rung plus O(rows) filters could replace k+1 full O(N)
  scans, a CPU win as well as a memory win. It needs a rows-restricted range kernel (only
  `polygon_select` takes `rows` today), i.e. a new C-ABI entry point and an `ABI_VERSION`
  bump, and it lands on the two already-red drill rows. Worth its own PR with its own A/B.
- **`to_pdf` inflates the SVG into an ElementTree**: 29.4 MB peak for a 1.07 MB PDF at only
  20k points, versus 5.8 MB for the same figure's SVG. Needs a streaming converter or
  conversion from the payload instead of from serialized SVG.
- **`to_html(path=...)` still materializes the whole document** before writing it; streaming
  parts to the file would remove one full copy but changes the return contract.
- Screen-grid mean-color worker accumulators cost 32 MB per colored figure (4 workers ×
  196k cells × 40 B); capping the fan-out at 2 halves it and costs wall-clock.
- The direct-tier raster scratch materializes xs/ys/radii/fills at 16 B/mark even when
  serial, partly because the fan-out estimate needs a full pass first.
- **`memory_report` should surface the transient high-water.** After a colored 2.1M drill
  session RSS is 382 MB while `resident_array_bytes` reports 120 MB; most of the gap is
  one-shot transients the allocator keeps rather than anything retained. §27's promise
  argues for an observed-RSS / peak-transient line.

## Summary by CodeRabbit

* **Performance**
  * Improved memory usage and speed for PNG, JPEG, SVG, and HTML exports, especially for large figures and images.
  * Reduced peak allocations during raster rendering, payload generation, and color/palette handling.
* **Enhancements**
  * Memory reports now include per-column capacity details and more accurate resident-memory totals.
  * Rasterization accepts a wider range of read-only buffer types without extra copies.
* **Bug Fixes**
  * Correct log-axis row rejection behavior in shipped payloads (with symlog unchanged) and improved handling for baseline-null dropping in area-like traces.
* **Documentation**
  * Updated the changelog and memory-model documentation accordingly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,15 @@ in the README).
     per-component token fields are freed as they are gathered. A 1800x840 export
     peaks at 48 MB instead of 108 (photographic: **400 MB instead of 1516 MB**)
     and is 5-19% faster.
+  - Standalone HTML export appends the base64 payload chunks as their own
+    document parts and joins once, instead of interpolating each chunk into a
+    wrapper string and folding a second joined copy into an f-string. The
+    encoded text is the bulk of the document, so that was two extra full copies
+    of it: a 1M-point export peaks at 34 MB instead of 41 and is marginally
+    faster.
+  - The native rasterizer's serial point and segment passes no longer
+    materialize an `0..n` index vector to read back in order (4 bytes per mark),
+    and the per-point density quantizer keeps one temporary instead of three.
 - `memory_report()` reports `capacity_bytes` per column and
   `canonical_capacity_bytes` per store, and builds `resident_array_bytes` from
   the capacity total. A streamed column's `values` is a prefix view of its
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,13 @@ in the README).
   - SVG documents assemble in one flat join with block-buffered markers, and the
     native rasterizer borrows the display list instead of freezing a `bytes`
     copy. A 100k-point SVG export peaks at 27 MB instead of 39 MB.
+  - The JPEG encoder streams instead of exploding: the entropy packer works in
+    bounded bit passes (it cost 17 bytes per output *bit*, so a 2.8 MB stream
+    peaked over 1.5 GB), the YCbCr planes are released as they are consumed, the
+    quantize chain rounds in place via `trunc(x + copysign(0.5, x))`, and the
+    per-component token fields are freed as they are gathered. A 1800x840 export
+    peaks at 48 MB instead of 108 (photographic: **400 MB instead of 1516 MB**)
+    and is 5-19% faster.
 - `memory_report()` reports `capacity_bytes` per column and
   `canonical_capacity_bytes` per store, and builds `resident_array_bytes` from
   the capacity total. A streamed column's `values` is a prefix view of its
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,34 @@ in the README).
   same validated style properties, so an explicit `style=` still wins and specs
   that don't use them are byte-identical.
 
+### Changed
+- Peak memory cut on four paths, with byte-identical output everywhere (89
+  payload/export/view fingerprints pinned before and after):
+  - The indexed-palette PNG encoder stages one scanline buffer and hands it to
+    zlib directly instead of building a per-row `bytes` list, joining it, and
+    narrowing an `intp`-per-pixel `np.unique` inverse. A 1800×840 export peaks
+    at 6 MB instead of 50 MB (33 MB instead of 274 MB at 4K) and encodes ~1.8x
+    faster; the truecolor branch drops ~17%.
+  - Full-column color quantization (density mean-color planes, `direct_rgba`
+    channels, u8 live-wire channels) runs chunk-bounded and in place. Its chunk
+    was 4M rows, so every real column still paid the one-shot peak: a 2.1M-row
+    continuous color channel resolved in 44 MB and now resolves in 7 MB, taking
+    a colored 2.1M-point first paint from 181 MB to 147 MB of RSS.
+  - Direct-tier scatter/line/area payloads skip the all-visible row mask. Zone
+    maps already count NaN *and* ±inf as null, so on linear axes with no nulls
+    the mask is provably all-true; it was three O(N) passes and two N-byte
+    temporaries per build. Emit is 16–35% faster, and area no longer allocates
+    an identity index vector (nor gathers animation keys through it).
+  - SVG documents assemble in one flat join with block-buffered markers, and the
+    native rasterizer borrows the display list instead of freezing a `bytes`
+    copy. A 100k-point SVG export peaks at 27 MB instead of 39 MB.
+- `memory_report()` reports `capacity_bytes` per column and
+  `canonical_capacity_bytes` per store, and builds `resident_array_bytes` from
+  the capacity total. A streamed column's `values` is a prefix view of its
+  capacity-doubling buffer, so up to half of what it holds was invisible to the
+  report (§27: if a number isn't in the report, it isn't real). Figures that
+  never appended report exactly what they did before.
+
 ### Fixed
 - The colorbar stringified its colormap, so a custom ramp reached it as an
   unparseable name and silently painted viridis while the marks beside it
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,12 +64,12 @@ in the README).
     per-component token fields are freed as they are gathered. A 1800x840 export
     peaks at 48 MB instead of 108 (photographic: **400 MB instead of 1516 MB**)
     and is 5-19% faster.
-  - Standalone HTML export appends the base64 payload chunks as their own
-    document parts and joins once, instead of interpolating each chunk into a
-    wrapper string and folding a second joined copy into an f-string. The
-    encoded text is the bulk of the document, so that was two extra full copies
-    of it: a 1M-point export peaks at 34 MB instead of 41 and is marginally
-    faster.
+  - Standalone HTML export joins the document once from parts, with every large
+    string — the client bundle, the spec, each base64 chunk — as its own part, so
+    the join copies it exactly once. Previously the chunks were folded through a
+    `"\n".join(...)` and then into an f-string, duplicating 4/3 of the payload. A
+    1M-point export peaks at 33 MB instead of 41; a small export (where the
+    ~330 KB client bundle is the document) is ~2% faster.
   - The native rasterizer's serial point and segment passes no longer
     materialize an `0..n` index vector to read back in order (4 bytes per mark),
     and the per-point density quantizer keeps one temporary instead of three.
--- a/python/xy/_figure.py
+++ b/python/xy/_figure.py
@@ -1871,8 +1871,10 @@ class Figure(AnnotationsMixin, PayloadMixin):
         report["transport_bytes_per_point"] = len(blob) / n_total
         report["pyramid_bytes"] = interaction.pyramid_report_bytes(self)
         report["bin_color_bytes"] = interaction.bin_color_cache_bytes(self)
+        # Capacity, not live length: a streamed column's growth-buffer slack is
+        # resident RAM (§27), and equals `canonical_bytes` when nothing appended.
         report["resident_array_bytes"] = (
-            report["canonical_bytes"]
+            report["canonical_capacity_bytes"]
             + report["channel_bytes"]
             + report["pyramid_bytes"]
             + report["bin_color_bytes"]
--- a/python/xy/_jpeg.py
+++ b/python/xy/_jpeg.py
@@ -20,7 +20,9 @@ export path must not grow an imaging dependency.
 
 from __future__ import annotations
 
+import itertools
 import struct
+from typing import Optional
 
 import numpy as np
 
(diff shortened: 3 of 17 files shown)


Alek99:
Addressed the review. Before writing anything I mutation-tested each ask, which changed what got done:

**The entropy-packer gap was real and worse than described.** Not only does no test cross a group boundary — I confirmed that dropping the carry, padding a non-final group, or skipping its byte stuffing each left the whole suite green. Now covered against a bit-by-bit statement of the T.81 rule: five tiny group sizes (so passes end at every byte alignment), the production group size over a stream crossing it twice, degenerate streams (empty, a single sub-byte token, consecutive `0xFF`), and end-to-end — a noisy 320×400 image encodes identically with the packer forced into one pass. All four failure modes above now fail the suite.

The tests deliberately pin the emitted bytes, not the grouping: packing is a concatenation with carries, so moving a split point by a token is output-equivalent. That mutation survives on purpose, and the test file says so, so nobody later reads it as a hole.

**The mask conditions: two of three were genuinely unpinned, and I checked rather than assumed.** Removing the log-axis condition, or the baseline-null condition, left all 2405 tests passing. Removing the null-rows condition failed `test_property_figure.py::test_area_hostile`, so that one was already covered. Added tests for the two that weren't: x and y log axes dropping non-positive rows, log combined with non-finite rows, symlog explicitly *keeping* non-positive rows, and an `error_band` whose baseline carries NaN/inf. Each was verified to fail against removal of the condition it covers.

Worth noting which direction is dangerous: weakening the predicate only recomputes a mask that rejects nothing (a wasted pass, not a wrong payload), while tightening it silently ships rows that should have been dropped. The new tests pin the tightening direction. The `symlog`-treated-as-`log` mutation therefore survives correctly.

**Report assertions** now cover the appended case — the case the change actually ships. Capacity exceeds live length, per-column capacity matches `Column.capacity_bytes`, `resident_array_bytes` is built from the capacity total, and growth stays inside the 2× amortized bound. Verified against two mutations (capacity ignoring `_grow`; `resident_array_bytes` reverting to live length).

**Both minor items taken.** `_log_visible_mask` now states the invariant it shares with `_visible_mask_needed` in both directions and points at the tests. `png_truecolor` takes `bytes | bytearray | memoryview | np.ndarray` instead of `object`, with the C-contiguity requirement spelled out — you were right that `object` was worse than the `bytes` it replaced.

`17960c1`: 2412 passed, and the 89 payload/export fingerprints plus 90 JPEG encodes are still byte-identical to `main`. Since this commit is tests plus comments plus one annotation, it cannot move the perf gate; the three A/B campaigns on the code commits stand.

One correction to the review's framing, for the record: `capacity_bytes` staleness via memmap isn't prevented only by the append-migration path — the report also never accumulates capacity in the memmapped branch, so both would have to break together.

Alek99:
Both items fixed in `84e0fc2`.

**CodSpeed's `test_html_export_line` −10.1% was a real regression, not runner noise.** That benchmark's document *is* the ~330 KB client bundle — a 100k-point line M4-decimates to a ~30 KB payload — and my parts list interpolated `client_js` into its header part, so the join copied the bundle a second time. It reproduces locally as +1.9% wall-clock on that exact shape, which is ~16 µs, i.e. one memcpy of 330 KB. That arithmetic is why I'm confident in the cause rather than just the correlation.

The assembler now follows one rule: any string that is large relative to the document is its own part and is never interpolated — the client bundle, the spec JSON, the inline decoder, and each base64 chunk. Each is then copied exactly once by the join, which is strictly fewer copies than `main` (main duplicated the base64 text; my first attempt traded that for duplicating the bundle).

After the fix, on the benchmark's shape: 0.730 → 0.726 ms best-of-three-runs, medians 0.792/0.803/0.846 → 0.785/0.787/0.794. A 1M-point export is 1.248 → 1.205 ms and peaks at 33.3 MB versus 34.0 before and 41.3 on main — so the fix improved both axes over my own earlier version.

Two notes on reading the rest of that report. The `⚠️ Different runtime environments` flag applies to the *improvements* too, and the two flagged there (`animation_plain_payload_100k` +20.7%, `animation_keyed_payload_100k` +13.0%) match my local interleaved A/B closely (−34.7% and −18.9% wall-clock), so I'd treat those as real. And the row I pre-emptively flagged in the PR description, `native_png_export_scatter`, came back untouched here — consistent with it having been local noise.

**@coderabbitai's `_gathered` note was correct** and I took it. A local stays alive until its frame returns, so `return np.concatenate(parts)[order]` held the pieces, the join, and the gather at once — three times the field size, not two, exactly as described. `parts` and `joined` are now dropped explicitly, so the docstring's claim is actually true.

One honest qualification: the measured peak on my probe images did not move (48.4 / 399.9 / 48.4 MB). Those images' binding peak is in `_component_tokens` construction rather than the gather, so this fix removes a real three-way coexistence that simply is not the high-water mark for them. It will matter for an image whose token count dominates, and the code no longer claims something it does not do.

All other 24 checks were green before this push.

PR #310, opened 2026-07-26
Mark style capabilities: stroke-linecap, marker-shape, mark plugins, and a support inventory

Alek99:
New customization capabilities, plus an honest inventory of which renderer draws what and how far each survives export.

## The rendering fix, shown

Line caps disagreed across the three renderers and none of them said so. Same chart, same 18px stroke, zoomed on the polyline's left end, no style property set — this is the default:

Line caps now agree across renderers

The area outline had the same bug one level quieter: the SVG writer named its join and let the cap default to the format's flat one, which `_pdf` then read back flat while the rasterizer drew round.

The area outline had the same bug

Both figures come from checking the pre-change tree into a worktree, building its Rust core and its client, and running the same figure through both — the only difference between columns is the change. Two facts fall out of the frames:

- **Native raster output is byte-identical before and after** (`np.array_equal`). `stroke_shaped` delegates to the original `stroke` for a round cap, so nothing that already renders moves.
- **Browser output changed in exactly 1024 pixels**, in two clusters at the polyline's two ends. That is the entire diff.

## What ships

**`stroke-linecap`** — `butt`/`round`/`square` on the line family, drawn identically by WebGL, SVG, and the native rasterizer.

The three stroke-linecap values

**`marker-shape`** — the CSS spelling of `symbol=`. Nearly free: all three renderers already drew 17 symbols and only the spelling was missing. Both forms build identical specs.

**Mark plugins** — `xy.register_mark(xy.MarkPlugin(...))` adds a chart kind xy does not ship; `xy.mark("name", ...)` uses it. A plugin supplies a `calc` over declared columns and a `build` returning *built-in* marks. Because its output is ordinary traces it reuses the built-in rendering, picking, and export paths rather than reimplementing them. `build` never sees the `Figure`, the trace list, or the column store. Declared columns arrive as canonical f64; the registry refuses to shadow a built-in kind, refuses to silently replace another plugin, and rejects column names `xy.mark()` binds itself.

**Export-styling contract** — `custom_css` was the only asymmetry the docs named, and it was already bounded: it raises. The unbounded one went unmentioned — `class_names` and per-slot `styles` are read only by the browser client, and the SVG writer emits no `class` at all, so a Tailwind-styled chart exports to PNG with none of it applied and no diagnostic. Now contracted in `spec/api/export.md` §9, the chrome-slots guide, and the limitations page, with `tests/test_export_style_survival.py` asserting the two native writers agree with each other and not merely with the page.

**Capability inventory** — `python/xy/styling/capabilities.py`, one entry per style property and chrome slot with its per-renderer support, generated into `spec/api/capability-matrix.md` and a public page. `tests/test_capability_registry.py` pins it to cover exactly `CHART_DOM_SLOTS` and exactly what `styles.py` compiles; which mark kinds accept a property is derived from `styles.py`, never typed out. I confirmed the guard bites by adding a fake property and watching the suite go red.

## Deliberately not here

**`stroke-linejoin` is not selectable.** Joins are always round — what the capsule distance field produced before caps became selectable. The WebGL client draws polylines as one instanced quad per segment with no join geometry, so a selectable join would be honored by two renderers and ignored by one. The native miter/bevel implementation and its wire byte are removed rather than shipped dark.

The SVG half stays, because it fixed a real defect: every stroked path now names `stroke-linejoin="round"` instead of letting the format's `miter` default through, which `_pdf` had been reading back as a mismatch with the rasterizer.

## Notes

`ABI_VERSION` 41 → 42: `OP_STROKE` and `OP_SMOOTH_STROKE` carry one extra cap byte.

`scripts/check_claim_guardrails.py` gains a plain ban on unqualifiable superlatives about the styling surface — "most customizable", "fully customizable", "style anything" — and now scans `README.md` and `CLAUDE.md`, which it previously did not. It caught two live overclaims in `docs/index.md` on its first run; both are rewritten.

## Verification

`cargo test` 119 passed · `cargo clippy -D warnings` clean · `uv run pytest` 2351 passed, 69 skipped · docs 91 passed · ruff, guardrails, generator `--check`, public-API, `abi_smoke` 133 checks, render smoke, append-stream smoke — all clean. No conflicts against `main` at 8edf1d0.

## Summary by CodeRabbit

* **New Features**
  * Added mark plugins, allowing custom marks to compose existing built-in marks.
  * Added `stroke-linecap` styling for line marks and `marker-shape` styling for scatter marks across supported renderers.
* **Documentation**
  * Added a capability matrix covering styling support, chart slots, renderers, and export behavior.
  * Added Custom Marks guidance, styling examples, and clearer export limitations.
* **Bug Fixes**
  * Improved line-cap rendering consistency across browser, SVG, and raster outputs.
  * Corrected SVG and raster output to honor configured stroke caps and marker shapes.


Alek99:
Worked through the review — 12 findings. Pushed `16bbfe0`. Eleven applied, one rejected with a test.

## The one that mattered

**The `fnmatch` rule-ordering bug was real and the most valuable finding here.** `*` spans dots and first-match wins, so `layout.*axis.showgrid` was claiming `layout.polar.angularaxis.showgrid` before the exclusion below it could reject it. 3-D, geo, polar, ternary, and smith axes were being counted as supported.

| | before | after |
|---|---|---|
| supported (whole schema) | 345 | **218** |
| mapped-with-difference | 126 | **115** |
| in-scope supported | 344 | **217** |

A 127-attribute overstatement — 37% — in the number this PR exists to make trustworthy. Corrected in the report, the dossier, the changelog, the skill, and the PR body. `test_excluded_subtrees_are_not_stolen_by_the_wildcard_axis_rules` pins the ordering so it can't silently regress.

**Second-worst: the stale "15 axis keys" in `xy-evaluate/SKILL.md`.** Exactly right, and worse than it looks — that file's entire job is to send agents to registries instead of stale prose, and it shipped with stale prose. Every count it states is now pinned to `capabilities.summary()`.

## Applied

`_plugin_column` → contiguous f64 before `calc` (§27) · reserved `xy.mark()` control-field names rejected at registration · `planned` rows labelled as *not accepted* in the generated table · "drawn by all three renderers" derived from `prop.support` instead of hardcoded · unsupported-reason table folds its tail and prints a total so the column sums · `KNOWN_RENDERER_DIVERGENCES` typed · guardrail now scans `CLAUDE.md`, the skill, and the comparison doc · `_emit_line` falls back instead of raising `KeyError` · dash values validated before both stroke walkers · PDF/JPEG/WebP added to the custom_css test · Altair pinned · line family enumerated in the docs · stray space and placeholderless f-string removed.

Two notes on those. The guardrail's comparative rule needed the same ±10-line window the numeric-multiplier rule already uses — a claim taxonomy states its scope above the table, and a sentence-level window rejected the very wording the document exists to model. And on dashes: I could not construct a hang (a negative entry advances the index rather than looping), so that change is hardening of a byte-decoder boundary, not a fix for a reproduced defect.

## Rejected: the miter limit

`mcos >= 1.0 / MITER_LIMIT` is correct; `1.0 / (2.0 * MITER_LIMIT)` would be wrong.

SVG defines the limit as a ratio of miter length to stroke **width**. With `phi` the turn angle between segment directions and `mcos = cos(phi/2)`:

- apex distance from the vertex = `hw / mcos`
- miter length = tip to inner corner = `2 * hw / mcos` = `w / mcos`
- ratio = `miterLength / w` = `1 / mcos`

So `ratio <= 4` ⟺ `mcos >= 1/4` = `1.0 / MITER_LIMIT`. The `2 *` would bevel at `mcos < 0.125`, admitting ratio-8 spikes — twice what SVG and PDF draw for the same path.

`miter_joins_bevel_past_svgs_four_to_one_limit` pins it at three angles (90° ratio 1.41 miters, 150° ratio 3.86 miters, 155° ratio 4.62 bevels). I applied the suggested value and confirmed the test fails on the third case, then reverted.

## Not doing: docstring coverage to 80%

Measured rather than assumed. My library code is now **100%** and my scripts **91%** — I added the missing ones. The gap is test functions, at 0%.

Repo baseline: `python/xy` 53.1%, `scripts` 18.7%, **`tests` 5.3%** (95 of 1,790). This repo documents tests with comments at the assertion, not docstrings, and my tests follow that — heavily. Adding 56 docstrings that restate adjacent comments would move a number without adding information, in a codebase that has never met that threshold anywhere. Happy to revisit if the 80% bar is one the project actually wants.

## Note on `cargo fmt`

Fixed the stray space, which was mine. I did not reformat the rest: `cargo fmt --check` already fails on `main` across `font.rs`, `kernels.rs`, `tiles.rs`, and `raster.rs`, and no workflow, Makefile target, or `verify_local` check runs it. Reformatting here would bury the diff in unrelated churn.

## Verification

`cargo test` 121 passed · `cargo clippy -D warnings` clean · `uv run pytest` 2370 passed · docs 91 passed · ruff, guardrails (84 documents), both generator `--check`s, public-API, `abi_smoke` 133 checks, render smoke — all clean.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #314, opened 2026-07-26
perf: tail-only f32 re-encode on streaming append

FarhanAliRaza:
## What

Streaming appends re-encoded 100% of every shipped geometry column per tick — an O(N) f64→f32 native pass plus an N-sized allocation per column — even though the sticky ship offset (`Column.suggest_offset`) keeps the encoding prefix-stable and the browser client already uploads only the tail (`bufferSubData`). This is item 1 of the 2026-07 Abseil perf audit.

This PR adds the kernel-side twin of the client's tail upload: a per-column whole-column f32 encode cache (`Column.encoded_f32`) served by an identity fast path in `_PayloadWriter.ship`, so each tick encodes only the appended rows.

## How

- Cache keyed on `(offset, scale, values base pointer)`, held in a capacity-doubling f32 buffer mirroring `Column.append`'s growth buffer.
- `Column.append` re-keys the cache across its own growth-buffer migrations (prefix preserved byte-for-byte) and drops it on external rebinding.
- Invalidation (offset recenter, scale change, shrink) re-encodes fully into a **fresh** buffer — previously shipped split-payload memoryviews (widget reopen state) are never overwritten.
- Filtered/decimated/masked ships (`values is not col.values`) bypass the cache unchanged.
- +4 B/point resident RAM per shipped column, itemized in `memory_report()` per column and totaled as `encode_cache_bytes` (§27: if it isn't in the report it isn't real).

## Numbers

- New CodSpeed benchmark `test_stream_scatter_append_direct` (direct tier pinned via `density=False`): **495µs → 205µs** per 1k-row append at 100k points.
- 8 direct traces × 199k points (append re-emits every trace): **8.96ms → 0.64ms** per tick (~14x).
- `test_stream_line_append` is expected to move less: decimated lines ship M4 views that bypass the cache by design.

## Tests

- `test_append_reencodes_only_the_tail` — spies on `kernels.encode_f32`; fails on main (4097-row full re-encode per 1-row append), passes here.
- `test_append_cached_encode_matches_full_reencode` — cached bytes are identical to a from-scratch encode.
- `test_shipped_buffers_survive_later_appends_and_recenter` — shipped zero-copy views stay byte-stable across tail extensions and cache invalidation.
- `test_memory_report_itemizes_encode_cache_bytes`.
- Existing prefix-stability suite (`test_append_keeps_offsets_sticky_and_prefixes_byte_identical` etc.) passes unchanged; `scripts/append_stream_smoke.py` passes end-to-end.

## Spec

- `spec/design-dossier.md` §27 memory rules: encode-cache corollary.
- `spec/design/wire-protocol.md` §4: kernel-side tail-only re-encode note.
- `spec/benchmarks/results.md`: `streaming_updates` row now inventories the CodSpeed streaming benchmarks.

## Summary by CodeRabbit

- **Performance**
  - Streamlined streaming appends by reusing warmed column encodings and encoding only newly appended rows when offsets/scales remain compatible.
  - Kept direct-tier behavior stable during tail-only updates, with trace point counts growing as expected.

- **Monitoring**
  - Memory reports now include encoding-cache usage (with per-column and total accounting).

- **Documentation**
  - Updated memory and append/wire-protocol specs plus benchmark documentation to reflect the new caching and tail-only encoding behavior.

- **Tests**
  - Added streaming append payload-prefix stability checks, cache accounting assertions, and a new CodSpeed direct-tier benchmark.


Alek99 (commented):
I found three issues in the current head (`1c5d68a`):

1. **High — rebinding can corrupt cached geometry** (`python/xy/columns.py:353-355`). When `_grow` exists and `values` was externally rebound, a cache warmed on the rebound array satisfies `cache[3] == old_ptr`. `append()` then restores the stale `_grow` prefix and re-keys the rebound encoding to that pointer. I reproduced a cached payload whose prefix differed from a fresh encode of `col.values`. Please confirm `values` still aliases `_grow`; otherwise rebuild growth storage from the current values and invalidate the cache.

2. **Medium — `resident_array_bytes` omits the encode cache** (`python/xy/_figure.py:1876-1881`). `encode_cache_bytes` is itemized but is not included in the combined resident total. For a 256-point scatter, the report returned 4096 canonical-capacity bytes, 2048 encode-cache bytes, and only 4096 resident-array bytes.

3. **Medium — the memory/complexity specification counts live points instead of allocated capacity** (`spec/design-dossier.md:997-1003`, `spec/design/wire-protocol.md:272-278`). The f32 buffer doubles capacity, so 257 live entries can retain 512 slots—nearly 8 B/live point, not +4 B/point. The per-tick O(appended rows) claim should also be described as amortized because a growth event copies the prefix.

Validation: `tests/test_streaming.py` passed (28 tests), and Ruff check/format passed. The broader figure/scatter run had 238 passes and one unrelated local Chromium screenshot failure.

masenf:
This seems like a non-trivial amount of extra memory usage to maintain the cache. At 4-8 B / point, we're talking hundreds of megabytes in large plots.

PR #316, opened 2026-07-26
Stop the PNG exporter from deleting text it cannot draw

Alek99:
Stacked on #315 (base: `alek/cust-audit-2`).

The native rasterizer bakes its glyphs into `src/font.rs` because the Rust core
deliberately avoids FreeType. A character outside that atlas got **no glyph and
no advance** — so it did not render as a box, it *vanished*.

before and after

Left is `to_png()` before: `Résumé` → `Rsum`, `Zürich` → `Zrich`,
`Köln` → `Kln`, the CJK tick is blank, and the `€` is gone from every y tick.
The same figure's `to_svg()` renders all of it correctly.

The currency case is not incidental — `y_axis(format="€,.0f")` shipped in #315,
and its symbol was being deleted from every tick of the exported PNG.

## Changes

- **Extend the atlas** to Latin-1 Supplement + Latin Extended-A (every language
  written in the Latin script) plus non-ASCII currency. `scripts/gen_font.py`
  now takes codepoint *ranges* rather than a hand-typed string, so the coverage
  is auditable. 205 → 424 glyphs.
- **Substitute U+FFFD** for anything still unbaked — CJK, Cyrillic, emoji —
  instead of dropping it. §28 asks that a decision the engine makes on the
  user's behalf be visible, and a replacement box is the visible form of "this
  renderer cannot draw that". Zero-width and control characters stay dropped;
  they have nothing to show.

## Cost

~48 KB of baked coverage: the core dylib goes 1,170,160 → 1,219,744 bytes
(+4.2%). Measured, not estimated.

CJK and emoji still need `engine=xy.Engine.chromium` for real glyphs — the
atlas approach cannot cover them, and this PR does not pretend otherwise. It
makes the limit visible instead of silent.

## Verification

- Pixel-level tests, not display-list tests: the dropping happens in Rust, so
  the Python tests compare `to_png()` bytes for `"X<char>"` against `"X"`.
  (My first probe instrumented `_Cmd.text` and wrongly showed everything
  passing — the display list carries the full string either way.)
- New Rust test pins the atlas rows and the U+FFFD fallback.
- 2462 Python tests and 119 Rust tests pass; ruff, ruff format, pre-commit
  clean; `ty` at the 13-diagnostic baseline; `abi_smoke` passes.
- `spec/api/styling.md` documented the silent-skip behavior precisely, so it is
  updated rather than left contradicting the code; `spec/design/rust-engine.md`
  glyph counts refreshed.

## Related, not in this PR

`tick_label_angle` at any angle other than ±90° is also discarded by this
exporter — labels are drawn horizontally and overflow the canvas, because the
display-list text command carries two rotation *flag bits* rather than an
angle. That is a wire-format change, so it gets its own PR.


Alek99:
Fixed in fdfac0db (on `alek/cust-audit-2`, merged forward into this branch and #318).

The regression was real and it was mine, from the base branch rather than this one: the probe that decides whether `color=` is a list of CSS colors called `arr.tolist()` unconditionally, ahead of `_factorize_categories` — which goes to some trouble to identify equal records in Rust precisely so it never creates N Python objects. Every categorical scatter paid for a full materialization it did not need.

A category column fails the colour test on its very first value, so the probe now gates on entry zero: `["setosa", ...]` costs O(1), and only an array that already looks like paint pays for the full scan. That is exactly as strict as the `all(...)` that follows, which already required every entry to be a colour string, so nothing changes classification.

Local timings at the benchmark's 100k rows, median of 12 runs:

| | median | min |
| --- | --- | --- |
| `origin/main` | 2.00 ms | 1.61 ms |
| this branch, after the fix | 2.03 ms | 1.64 ms |

The probe itself went from 7.9 ms to 30 µs on a 500k-row category column.

Pinned with a test rather than a timing assertion — an `ndarray` subclass whose `tolist` raises, so the gate cannot silently regress.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #348, opened 2026-07-27
Give every capture-owning gesture the same capture-loss policy

### LINKED ISSUE #212: when the cursor leaves the chart area, sometimes the pan state sticks
attempt to pan a chart, have the mouse leave the chart area then release the mouse button. when the mouse goes back into the chart area it acts like the button is still clicked and continue to pan the chart.

https://github.com/user-attachments/assets/ca88d10e-cf3a-4ab6-b5ac-343bdd26bcab

FarhanAliRaza:
Fixes #212.

## The bug

Pan a chart, move the cursor out of the chart area, release the button outside
it — on returning, the chart keeps panning as though the button were still
held.

Pointer capture is scoped to a single browsing context. When a drag leaves an
embedded iframe and the primary mouse button is released in the parent
document, the iframe never receives `pointerup`; Chrome reports the lost
capture only once the pointer re-enters. The gesture stayed live across that
gap, so the next `pointermove` resumed it.

The plot canvas is where this is reported, but capture is acquired in **five**
places, and the axis-band drag, the lasso vertex handles and the modebar drag
all had identical exposure. Fixing only the pan would have left the same bug
reachable three other ways.

## The fix

Capture is now acquired in exactly one place.
`ChartView._captureGesturePointer` owns the detector — `lostpointercapture`,
plus a buttonless mouse move as the backstop for browsers that delay or omit it
— and each gesture declares its own end policy:

| Gesture | On capture loss |
| --- | --- |
| Canvas pan | Finalizes at its last in-frame view, emits its one `end` event with `interaction_id` intact, so history still coalesces the gesture into a single entry |
| Canvas select / box-zoom | Rolls back to the previous selection — the release coordinate that would complete it never arrived |
| Lasso vertex handle | Restores the prior vertex |
| Axis-band drag | Pan finalizes; a span-zoom rolls back |
| Modebar drag | Ends at its last in-frame position |

Re-entering the chart never resumes a gesture.

### Listener lifecycle

The helper registers through the existing `_listen` registry rather than a
private `addEventListener`, because two subsystems depend on that registry
being complete: `destroy()` sweeps it, and WebGL context-loss recovery walks it
to re-bind canvas handlers onto the replacement canvas. `_unlisten` detaches
via the record's *live* target, so a handler the canvas swap moved still
detaches from the node it ended up on.

Escape also now releases before dropping its gesture records. Previously it
nulled them while the capture object was the only handle on its listener,
stranding a listener that stayed armed and re-fired on later gestures.

## Scope limit, recorded

The buttonless-move backstop is **mouse-only** — it reads the primary-button
bit, which has no pen/touch equivalent. Pen and touch rely on
`lostpointercapture` alone. Recorded in
`spec/design/pan-and-zoom-configuration.md` rather than left implicit.

## Coverage

`test_capture_loss_policy_covers_every_capture_owning_gesture` drives all five
gestures through a capture loss in headless Chromium and asserts each one's end
policy.

The source-level guard in `test_static_client_security.py` asserts the durable
invariant — exactly one `setPointerCapture` site, so capture-loss handling
cannot be opted out of. Which gestures inherit the policy is asserted
behaviorally by the probe rather than by grepping for a call count.

## Specs

Updated per the repo's spec-as-source-of-truth rule:
`spec/api/interaction.md` (observable contract), and the mechanism plus its
mouse-only scope in `spec/design/pan-and-zoom-configuration.md` and
`spec/design/view-state.md`.

## Verification

- `node js/build.mjs` — typechecks and rebuilds
- `uv run pytest` — 3187 passed, 1 skipped
- `scripts/abi_smoke.py`, `scripts/render_smoke_nonumpy.py` — pass (the render
  smoke exercises context-loss canvas swaps and `destroy()`, the paths
  `_unlisten` touches)
- `pre-commit run --all-files`, `ruff check .`, `ruff format --check .` — clean

## Note for reviewers

Two things I deliberately did not do, both worth their own change:

- **A fuller `_beginGesture(owner, event, {onMove, onEnd(reason)})`** that owns
  `pointerup`/`pointercancel` too. Today the helper centralizes the *detector*;
  each gesture still hand-writes its `onLost` and its own `release()` in the
  normal end path, so a sixth gesture could still forget one. That would be a
  redesign of all five gestures.
- **Splitting the ~400-line `_initInteraction` closure** into per-gesture
  binders.

Separately: `tests/reflex_adapter/test_assets.py::test_client_source_is_the_installed_bundle`
fails locally because it expects unminified markers in `static/index.js` while
`js/build.mjs` emits a minified bundle. It is **pre-existing** — it fails
identically with this branch's changes stashed — and untouched here.

### DIFF
--- a/js/src/50_chartview.ts
+++ b/js/src/50_chartview.ts
@@ -806,6 +806,51 @@ export class ChartView {
     return handler;
   }
 
+  // Detach a handler registered through `_listen`. The record carries the live
+  // target, so a listener that context-loss recovery re-bound onto a
+  // replacement canvas still detaches from the node it ended up on.
+  _unlisten(handler) {
+    const index = this._listeners.findIndex((record) => record.handler === handler);
+    if (index === -1) return;
+    const [record] = this._listeners.splice(index, 1);
+    record.target.removeEventListener(record.type, record.handler, record.options);
+  }
+
+  _captureGesturePointer(owner, event, onLost) {
+    const pointerId = event.pointerId;
+    let active = true;
+    const release = () => {
+      if (!active) return;
+      active = false;
+      this._unlisten(lost);
+      // Guarded, so this is a no-op when the browser already took capture back
+      // (a real `lostpointercapture`, or the implicit release after pointerup).
+      try {
+        if (owner.hasPointerCapture(pointerId)) owner.releasePointerCapture(pointerId);
+      } catch (_err) { /* synthetic event */ }
+    };
+    const lost = (lostEvent) => {
+      if (!active || lostEvent.pointerId !== pointerId) return;
+      release();
+      onLost(lostEvent);
+    };
+    const guard = (moveEvent) => {
+      if (!active || moveEvent.pointerId !== pointerId) return false;
+      // Pointer capture cannot cross a browsing-context boundary. A mouse
+      // released outside an iframe can return without pointerup; treat the
+      // first trusted buttonless move as the missing capture-loss signal.
+      if (moveEvent.type === "pointermove" && moveEvent.isTrusted
+          && moveEvent.pointerType === "mouse" && !(moveEvent.buttons & 1)) {
+        lost(moveEvent);
+        return false;
+      }
+      return true;
+    };
+    this._listen(owner, "lostpointercapture", lost);
+    try { owner.setPointerCapture(pointerId); } catch (_err) { /* synthetic event */ }
+    return { guard, release };
+  }
+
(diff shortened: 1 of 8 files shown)


masenf:
> The deeper shape is one capture-loss policy every
capture-owning gesture inherits; that is a larger change than this fix.

It seems this is the fix we want anyway though. Seems weird to half-fix the issue just for panning.

masenf (approved):

PR #368, opened 2026-07-28
Remove unused _pow10_label helper from pyplot axes

aymuos15:
`_pow10_label` in `python/xy/pyplot/_axes.py` was a one-line wrapper over `_pow_label(value, 10.0)` with no call sites anywhere in `python/`, `tests/`, `scripts/`, `js/`, or `docs/`. 

`Axes._apply_tickers` already calls `_pow_label` directly with the base pulled from the tick spec (`_pow_label(value, float(spec.get("base", 10.0)))`), so the hardcoded-base wrapper had been superseded.


aymuos15:
Seems to be another flaky case.

PR #383, opened 2026-07-30
Make Binder build the native Rust core

### LINKED ISSUE #379: examples cannot be run on online jupyter notebook (mybinder) due to Rust core missing
reproducible via https://hub.bids.mybinder.org/user/reflex-dev-xy-igq1vmkc/doc/tree/examples/real_world/04_dukascopy_fx_ticks.ipynb

```error
File [/srv/conda/envs/notebook/lib/python3.14/site-packages/xy/kernels.py:18](https://hub.bids.mybinder.org/srv/conda/envs/notebook/lib/python3.14/site-packages/xy/kernels.py#line=17)
     16     from . import _native as _impl
     17 except ImportError as err:  # pragma: no cover - platform-dependent
---> 18     raise ImportError(
     19         "xy requires its native Rust core, which could not be loaded "
     20         f"({err}). Prebuilt wheels cover Linux glibc and musl (x86-64, aarch64, "
     21         "armv7), macOS (x86-64, Apple Silicon), and Windows (x86, x64, arm64); "
     22         "on those platforms `pip install xy` needs no toolchain. On any "
     23         "other platform, install a Rust toolchain (https://rustup.rs/) and "
     24         "reinstall from source (or run `cargo build --release`)."
     25     ) from err
     27 BACKEND = "native"
     29 CSS_DECLARATION = _impl.CSS_DECLARATION

ImportError: xy requires its native Rust core, which could not be loaded (xy native core not found (looked for libxy_core.so in ['[/srv/conda/envs/notebook/lib/python3.14/site-packages/xy/_native_lib/libxy_core.so](https://hub.bids.mybinder.org/srv/conda/envs/notebook/lib/python3.14/site-packages/xy/_native_lib/libxy_core.so)', '[/srv/conda/envs/notebook/lib/python3.14/target/release/libxy_core.so](https://hub.bids.mybinder.org/srv/conda/envs/notebook/lib/python3.14/target/release/libxy_core.so)', '[/srv/conda/envs/notebook/lib/python3.14/target/debug/libxy_core.so](https://hub.bids.mybinder.org/srv/conda/envs/notebook/lib/python3.14/target/debug/libxy_core.so)']). No prebuilt wheel exists for this platform — see the xy README for supported platforms, or build from source with `cargo build --release`.). Prebuilt wheels cover Linux glibc and musl (x86-64, aarch64, armv7), macOS (x86-64, Apple Silicon), and Windows (x86, x64, arm64); on those platforms `pip install xy` needs no toolchain. On any other platform, install a Rust toolchain (https://rustup.rs/) and reinstall from source (or run `cargo build --release`).
```

FarhanAliRaza:
### Motivation

- Binder-built images were installing the repository checkout as a pure-Python package when `cargo` was absent, leaving notebooks that fail at import time because the native Rust core is missing (#379).
- The goal is to make hosted Binder images fail early (during image build) when the native core cannot be produced, and to make the examples actually runnable — most notebooks need a scientific stack that neither the checkout install nor repo2docker's base environment provides.

### Trade-off, stated explicitly

Binder is linux-64, where published wheels already exist — `pip install xy` would need no toolchain and carry no build risk. This PR instead takes on a full `cargo build --release` + `npm ci` + vite build inside repo2docker (build-timeout and image-size exposure) deliberately, so the notebooks always run against the launched ref rather than the last release. **Fallback if the source build breaks:** swap the `postBuild` install to `pip install xy` (published wheel), losing only launched-ref fidelity. This is recorded in `spec/design-dossier.md`.

### Description

- `.binder/environment.yml` provisions only the build toolchain with loose, arch-agnostic pins (`rust=1.88.*`, `nodejs=22.*`). Exact conda build strings are deliberately avoided: they are arch-specific (mybinder also schedules aarch64 nodes) and conda-forge rebuilds retire them, which silently rots the file into an unsolvable image.
- The interpreter carries a loose version-only pin (`python=3.13.*`): repo2docker's default kernel env resolves to Python 3.10 (observed 3.10.19 on both mybinder.org and CI), below the package's `requires-python >=3.11`, so without the pin the pip install fails at metadata. pip stays unpinned.
- The scientific stack imported across `examples/**/*.ipynb` (matplotlib, pandas, scipy, scikit-learn, seaborn, h5py, requests, plus pysam and gwosc for the `real_world` series) installs from pip wheels on the `postBuild` install line rather than through conda — keeping ~200 packages out of the mamba solve makes the build faster and sidesteps a reproducible mamba extraction failure observed on a mybinder builder node ("Cannot find a valid extracted directory cache").
- `.binder/postBuild` (executable) exports `XY_REQUIRE_CARGO=1` so a missing toolchain fails the image build instead of shipping a coreless wheel, disables Playwright's browser download (`PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1`), installs the checkout + notebook stack with `pip --no-cache-dir`, and removes the build-only caches (`node_modules`, `target`, cargo's crate downloads, npm's cache).
- `tests/test_binder_config.py` asserts **properties** of the parsed config rather than transcribing its contents: contract toolchain versions present, no arch-specific build strings, interpreter pinned at or above the `requires-python` floor, both env exports ordered before the pip install, notebook stack on the install line, cache removal after it, script executable (skipped on Windows checkouts). It `importorskip`s PyYAML so the python-floor CI job (bare package + pytest by design) skips it cleanly.
- New `.github/workflows/binder.yml` runs a pinned `jupyter-repo2docker==2026.4.0 --no-run .` on changes to `.binder/**`, `pyproject.toml`, or `hatch_build.py` — nothing else in CI builds a Binder image, so this is the only check that exercises the config end to end (it caught the Python-floor conflict on this very PR).
- README gains a Binder launch badge and `docs/integrations/notebooks.md` a "Run the Examples on Binder" section, so the deployment is discoverable. Both are explicit that the badge launches `main`.
- Binder source-build contract, pin policy, and fallback documented in `spec/design-dossier.md`.

### Testing

- End to end on mybinder.org: the branch image built and launched, and the #379 notebook (`04_dukascopy_fx_ticks.ipynb`) opens with the native core present.
- `.github/workflows/binder.yml` (full `repo2docker` image build) green on this PR.
- `uv run pytest -q tests/test_binder_config.py`, pre-commit, ruff check/format — all pass; the missing-PyYAML floor environment was simulated locally and skips cleanly.

------
Fixes https://github.com/reflex-dev/xy/issues/379


Alek99 (changes requested):
The diagnosis and the mechanism are right — `.binder/` + conda `rust`/`nodejs` + a source install under `XY_REQUIRE_CARGO=1` is exactly how you make repo2docker produce a native core. But as written **the Binder image cannot build**: three of the four conda pins do not exist on conda-forge, so the environment solve fails before `postBuild` ever runs. And with the solve fixed, most `examples/` notebooks still won't run.

Green CI is not evidence here — nothing in CI builds a Binder image.

## 1. Blocking: 3 of 4 conda pins don't exist

Checked each pinned dist against anaconda.org (method validated with a known-good control dist that returns 200):

| pin in `.binder/environment.yml` | exists? | reality on conda-forge |
|---|---|---|
| `python=3.13.5=h2b335a9_102_cp313` | **404** | the linux-64 build is `hec9711d_102_cp313` (`h2b335a9` is the 3.13.**8** hash) |
| `nodejs=22.16.0=hb8e1007_0` | **404** | **22.16.0 was never packaged** — linux-64 goes 22.13.0 → 22.17.0 |
| `rust=1.88.0=h3a20983_0` | **404** | the linux-64 build is `h1a8d7c4_0` |
| `pip=25.1.1=pyh8b19718_0` | ✅ noarch | — |

Net effect: `mamba env update` fails with "nothing provides …", the image build dies, and #379 gets *worse* — a visitor sees an opaque build log instead of today's actionable `ImportError`.

## 2. Blocking for the issue's goal: notebook dependencies are missing

`pip install .` brings only `numpy` + `anywidget`. repo2docker's frozen 3.13 base env has none of the scientific stack (its lock carries only `matplotlib-inline`, plus `requests`). Imports across `examples/**/*.ipynb`:

```
38 numpy   20 matplotlib   10 sklearn   8 requests   6 pandas
 3 seaborn   2 scipy   2 pysam   2 urllib3   1 h5py   1 gwosc
```

The specific notebook in #379 (`04_dukascopy_fx_ticks`) survives because `requests`/`urllib3` happen to be in the base — but the pdsh series, `matplotlib_shim.ipynb`, and everything touching sklearn/pandas/seaborn/scipy still fail at import. #379 is "examples cannot be run on online jupyter notebook"; fixing the `xy` import alone doesn't close it. Please add the notebook deps to `environment.yml` (matplotlib, pandas, scipy, scikit-learn, seaborn, h5py at minimum) and either add or explicitly scope out the exotic ones (`pysam`, `gwosc`).

## 3. Pinning strategy, once the strings are corrected

- **Exact build strings are the wrong granularity here.** They are arch-specific (mybinder also runs aarch64 nodes) and conda-forge repackages/removes builds, so this file will silently rot back into an unbuildable image.
- **Pinning Python *below* repo2docker's base is actively harmful.** Its 3.13 lock ships `python 3.13.8` and `pip 25.2`; asking for 3.13.5 / pip 25.1.1 forces mamba to downgrade the interpreter and churn the whole preinstalled Jupyter stack.
- The wheel is ABI-agnostic (`py3-none-<plat>`, ctypes C ABI), so **nothing here needs a Python pin at all** — the 3.14 default that reported the bug would work fine. Suggest dropping `python`/`pip` and using `nodejs=22.*`, `rust=1.88.*`.
- The inline comment says 3.13 is "covered by XY's test matrix", but CI's only explicit pin is `3.11` (the floor job) and there is no `.python-version`. The stated justification doesn't hold.

## 4. The test locks in the wrong values, and can't detect a structure error

Running the test body on this branch passes — with unbuildable pins. That's the problem: it asserts transcription, not solvability.

- The `dependencies` set is built from *every* `  - ` line, so `conda-forge` (a **channel**) is asserted as a dependency (`tests/test_binder_config.py:18`). Swap the `channels:` and `dependencies:` keys and the test still passes.
- Exact `commands ==` matching means any added comment or reorder in `postBuild` breaks the test, while a semantically broken script that keeps the shape passes it.
- `not line.startswith("#!")` is dead code — already covered by `startswith("#")`.
- `stat().st_mode & 0o111` fails on a Windows checkout (CI's pytest is ubuntu-only, so this one is only a local-dev nit).

Better: parse the YAML properly and assert *properties* — `rust` and `nodejs` present, `XY_REQUIRE_CARGO=1` exported before the `pip install`, the Playwright download disabled, the script executable. The only check with real signal would be a CI job running `jupyter-repo2docker --no-run .`, which would have caught finding 1 outright. Probably worth the runner minutes, since nothing else in the repo exercises this config.

## 5. Worth stating in the PR description

Binder is linux-64, where published wheels already exist — `pip install xy` needs no toolchain and carries no build risk. This PR instead takes on a full `cargo build --release` + `npm ci` + vite build inside repo2docker (build-timeout and image-size exposure) purely to keep notebooks on the browsed commit. That's a defensible trade-off, but it should be stated explicitly, ideally with a documented fallback for when the source build fails.

## What's correct

- `XY_REQUIRE_CARGO=1` is the right lever — `hatch_build.py:171`. Without it the hook silently ships a pure-Python wheel, which is precisely the #379 failure mode.
- `nodejs` is genuinely required, not incidental: the JS client is a hard requirement of every distribution and is built with `npm ci && node js/build.mjs` from a clone.
- `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1` is the correct variable and it matters — `playwright` is in `devDependencies`, so `npm ci` would otherwise pull several hundred MB of browsers into the image.
- `rm -rf node_modules target` is right for image size; the installed wheel force-includes the lib into site-packages, so nothing at runtime depends on those directories.
- `.binder/postBuild` is committed executable (100755), and there is no competing root-level repo2docker config to conflict with `.binder/`.

One aside: the repo has no Binder references at all (no launch badge, no docs link), so once this works nothing points users at it.

PR #452, opened 2026-08-04
Add a report-only style-compatibility preflight and the applicable-slot matrix

Alek99:
First change of the styling-compatibility program (Phase 0 of the plan): make every export explain itself before changing anything about how exports behave.

## What this adds

- **`chart.style_compatibility_report(target=..., engine=..., custom_css=...)`** — a report-only preflight that routes every declared slot style for one export target into exactly one of four outcomes: `survives`, `native-subset` (naming the kept and lost properties per format family), `browser-only`, or `state-gated`. It mirrors the export path's own refusals (`custom_css` with a pinned native engine, Chromium SVG) by calling the same resolver, so it can never disagree with what running the export would do. Report-only: computing it changes no export. The staged `compatibility=` modes that act on the report come in the follow-up PR.
- **The applicable-slot matrix.** The capability registry now tags each of the 48 chrome slots with an applicability — present in a clean static export (24), or gated by an export state: `hover` (5), `modebar` (14), `crosshair` (2), `selection` (1), `view`/reduction badges (2). The generated matrices gain the *applicable in* column and applicable-slot counts. A styled tooltip is no longer counted as "dropped" by a static PNG that never contains a tooltip — recorded with its gating state instead (§28: the distinction is written down, not silent).

## Contract properties (tested)

- **Constant-time when there is nothing to route**: no `class_names`, no per-slot `styles`, no `custom_css` → no slot walk. Preflight stays free exactly where exports are hot.
- **One source of truth per fact**: routing derives from the capability registry, honored property subsets from the writers' own constants (`_svg.SLOT_TEXT_PROPS` / `SLOT_RASTER_PROPS`), engine selection from `export._resolve_image_engine`. The preflight restates none of them.
- **Partition is pinned**: new registry tests fail if a future `modebar_*`/`tooltip*`/`crosshair_*`/`badge*` slot forgets its family state, and reject native support claims on state-gated slots until the interaction-snapshot phase adds them deliberately.
- The legend slot is qualified rather than guessed in either direction (its box properties route through the merged legend declaration, which has no constant yet — the shared-IR PR makes it exact).

## Evidence

- 20 new preflight tests + 4 new registry tests; full suite: **3973 passed, 109 skipped**.
- `spec/api/export.md` §9 gains the report as its programmatic form; `spec/api/capability-matrix.md` and `docs/styling/capabilities.md` regenerated (the `--check` drift gate covers both).
- `ruff check` / `ruff format --check` / pre-commit: clean. `ty check`: no diagnostics in the new code (baseline count unchanged).

Part of the full styling-compatibility plan (resolved-style IR → live snapshot → native chrome parity); next in the stack: opt-in `compatibility=` warn/strict modes and the engine-collision contract.

<a href="https://cubic.dev/pr/reflex-dev/xy/pull/452?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added `style_compatibility_report()` for charts and figures to preview how styling will behave for a selected export format and engine.
  * Reports identify styles that survive export, are partially supported, browser-only, or dependent on interaction and view state.
  * Export compatibility checks now include custom CSS and clearly explain unsupported styles and export refusals.

* **Documentation**
  * Updated capability references to distinguish static-export features from state-dependent chart features, including applicability details for all styling slots.


Alek99:

PR #453, opened 2026-08-04
Stage the styling contract: compatibility= modes and the engine collision table

Alek99:
**Stacked on #452** (preflight report + applicable-slot matrix) — second change of the styling-compatibility program's Phase 0.

## What this adds

Every image-export API (`to_png`, `to_svg`, `to_image`, `write_image`, batch `write_images` — at chart, figure, and module level) accepts `compatibility=`:

| Mode | Behavior |
|---|---|
| `"legacy"` (default) | Exactly today's behavior — byte-identical output, tested. |
| `"warn"` | One `StyleCompatibilityWarning` naming each declaration the export would drop; bytes still produced. |
| `"strict"` | `StyleCompatibilityError` **before emission**, full preflight report attached, remedies in the message. |
| `"lossless"` | **Reserved and rejected** until `Engine.auto` can actually re-route on preflight evidence — accepting the name earlier would make it a lie. |

## The engine contract

Engine selection and compatibility are orthogonal; an explicit engine is a hard constraint no mode may override:
- strict + pinned native engine → refuses on that engine, never silently re-routes to Chromium;
- any mode + Chromium engine → the browser renders the full cascade, nothing can drop, the mode has nothing to do (tested against a live Chromium);
- resolution errors (`custom_css` + pinned native, Chromium SVG) precede and outrank mode logic in every mode — today's `ValueError`s fire unchanged;
- a strict batch fails whole while the plan is resolved, before any file is written;
- HTML rejects a non-default mode like the other options it cannot honor.

State-gated chrome never trips warn/strict in a clean static export (the applicable-slot contract from #452, now enforced rather than only reported).

## Performance posture

- The default path pays **one string comparison**: literal-`legacy` returns before the preflight machinery is imported.
- `xy.export`'s `StyleCompatibilityError`/`Warning` aliases resolve via module `__getattr__`, so importing the module stays exactly as heavy as before — verified that the eager alias would have pulled the native dylib into an import chain that never had it, and rejected for that reason.
- warn/strict pay the routing walk only for charts carrying class or per-slot declarations on a native engine.

## Migration schedule (named, per Phase 0's rule)

`spec/process/style-compatibility-migration.md`: `compatibility=` ships default-`legacy` in 0.0.6 → default flips to `warn` in 0.0.7 → `strict` at the 0.1.0 minor boundary → `legacy` **removed in 0.2.0**. Facet grids deliberately don't accept the option yet (tracked there) rather than half-honoring it.

## Evidence

14 new mode tests covering the engine × mode matrix (including live-Chromium runs); the one adjusted existing test is the Chart→Figure delegation contract, extended for the grown kwarg. Full suite: **3987 passed, 109 skipped**. ruff/format/pre-commit clean; `xy.export` import weight verified unchanged.

<a href="https://cubic.dev/pr/reflex-dev/xy/pull/453?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Alek99:

PR #454, opened 2026-08-04
Land the ResolvedStyleSnapshot: schema v1, interned, with a generated TS mirror

Alek99:
**Stacked on #453** (compatibility modes), which stacks on #452 (preflight) — third change of the styling-compatibility program, and the schema every later phase consumes (slot migration, browser capture, native cascade).

## What this adds

`python/xy/styling/resolved.py` — the renderer-neutral styling IR — plus its generated TypeScript mirror `js/src/14_style_snapshot.ts` (`scripts/gen_style_snapshot_types.py`, with `--check` wired into the suite so the two languages cannot drift; the `gen_capability_matrix.py` contract applied to types).

Three contract properties, enforced at construction on both ends of the eventual wire:

- **Concrete values only.** No `var()`/`calc()`/`env()`/`inherit`, no relative units (`em`, `%`, `vw`, …). A value that still depends on a cascade — or on metrics the consumer would re-derive — is rejected with the reason (§28). One smuggled unresolved value would re-create inside the IR exactly the per-renderer divergence the IR exists to end.
- **Interned declarations.** Each distinct declaration is stored once; instances reference it by index carrying only identity qualifiers (`["y","major","3"]`), resolved geometry, and content. Interning is canonicalized (insertion-order independent). The dense-axis fixture — 460 instances, 2 declarations — serializes to **~38.6 KB against the spec's 50 KB budget**, asserted with the headroom stated so eating it is a decision, not drift.
- **Closed vocabulary per version.** Schema v1's property groups (paint / typography / layout / effects) are generated constants in both languages; growing the vocabulary is a `STYLE_SNAPSHOT_VERSION` bump. `snapshot_from_payload` refuses unknown versions rather than guessing — a snapshot is cacheable and supplyable across builds, so version humility is load-bearing.

## Protocol posture

Nothing rides the wire yet: `PROTOCOL_VERSION` stays at 12. `spec/design/wire-protocol.md` gains §8 documenting the payload shape and **reserving** `style_snapshot_request` / `style_snapshot` for the capture change, which bumps the protocol and carries this schema as its reply. The client build typechecks the mirror; nothing imports it yet.

## Evidence

25 new schema tests (interning, order-independence, concreteness rejections with reasons, wire round-trip, malformed-payload refusals, mirror freshness). Full suite: **4012 passed, 109 skipped**. `node js/build.mjs` typechecks + builds clean; ruff/format/pre-commit clean.

Next in the stack: the existing ten native-capable slots routed through this IR behind the `_compile_cached` style cache, gated byte/pixel-equivalent.

<a href="https://cubic.dev/pr/reflex-dev/xy/pull/454?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Alek99:

PR #467, opened 2026-08-05
Data-bound chart components 7/8: demo app on the new API

FarhanAliRaza:
Stacked on #466. **Base is `stack/6-compile-probe`.** Top of the stack — check this branch out to see the whole feature.

## Change

Ports `examples/reflex` to the tier the stack built, so the showcase is the API we recommend rather than the one it replaces:

- a composed 1M drillable scatter with hover/click/select;
- an `on_view_change` data var republishing in-view columns into a fixed histogram plan;
- a flat scatter whose slider republishes columns under a stable handle;
- an `rx.cond` toggle between a composed board and `rx.foreach` small multiples over a `list[DataHandle]` var.

**The escape hatch stays represented on purpose.** The cross-filtered histogram whose *structure* reads state keeps `@reflex_xy.figure`, alongside both fixed-data tiers and the FastAPI cross-host A/B — so the example shows where each tier is the right answer rather than implying the new one subsumes them.

`reflex_ws_smoke.py` drives the ported app; `test_example_apps.py` compiles it.

## Spec

`reflex-integration.md` file map (examples/reflex inventory).

## Test plan

- `uv run pytest tests/reflex_adapter tests/test_validation_timing.py tests/test_example_apps.py` — 250 passed, 1 skipped
- `pre-commit run --all-files`, `ruff check`, `ruff format --check`, `ty check` — clean
- Browser E2E (`scripts/reflex_ws_smoke.py`) against the running demo not re-run in this session — worth doing before merge.

<a href="https://cubic.dev/pr/reflex-dev/xy/pull/467?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,6 +250,44 @@ app = rx.App()
 app.add_page(index)
 ```
 
+For state-driven charts, declare the chart in the page and supply columns from
+a `@reflex_xy.data` state method — the structure (channels, colormaps, axes)
+is validated when `reflex run` compiles the app, while the columns ride the
+app's own websocket as binary buffers, never through Reflex state:
+
+```python
+from typing import TypedDict
+
+import numpy as np
+import reflex as rx
+import reflex_xy
+
+
+class CloudData(TypedDict):
+    x: np.ndarray
+    y: np.ndarray
+    mag: np.ndarray
+
+
+class Dash(rx.State):
+    points: int = 200_000
+
+    @reflex_xy.data
+    def cloud(self) -> CloudData:
+        rng = np.random.default_rng(7)
+        x = rng.normal(size=self.points)
+        y = x * 0.6 + rng.normal(scale=0.6, size=self.points)
+        return {"x": x, "y": y, "mag": np.hypot(x, y)}
+
+
+def dashboard() -> rx.Component:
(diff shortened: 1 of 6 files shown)


FarhanAliRaza:
Review addressed in 24569d2c: the §2 `set_bins` and §8 `set_bound_points` events now validate the payload's list shape and clamp the number to the slider's min/max server-side (`_clamped_slider_value`) before assignment — slider limits are UI hints, not a security boundary, and both values size server allocations. The module docstring's section count is corrected.

Also, the /kinds page (stacked on top in 8/8) now has **browser render coverage**: `scripts/reflex_ws_smoke.py` step 7 navigates there and pixel-probes all 26 kind cells; the full smoke passes end to end against the running app (all 26 cells paint, one shared websocket, drill/stream/cond steps green).

PR #476, opened 2026-08-07
Use native factorization for object string categories

### LINKED ISSUE #471: Use the native factorizer for low-cardinality object categorical channels
## Summary

Object-dtype categorical channels fall back to a Python pass over every row in `_factorize_categories`, even when the column is a low-cardinality repeated category column that could use the existing native factorizer. This creates an avoidable backend performance cliff for common pandas object/string columns.

## Current behavior

In `python/xy/channels.py`, `_factorize_categories()` uses the native fixed-width factorizer only for `U`, `S`, and boolean NumPy dtypes. Object arrays always execute:

```python
labels = [category_label(value) for value in arr.astype(object)]
categories = sorted(set(labels))
```

For a large object column with a small repeated category set, this materializes an N-entry Python label list and performs Python-level mapping despite the data being suitable for native equality/factorization after a safe normalization step.

The performance audit identifies this as a significant cost for common object-backed pandas string/category columns: the current path performs multiple Python-level passes and retains an N-entry label list during payload construction.

## Proposed direction

Add a semantics-preserving object-array fast path for values that can be normalized to a fixed-width native representation, then route those values through the existing native factorizer. Retain the current Python fallback for mixed objects, null-like values, custom `category_label` cases, and any values whose canonical display labels can collide or change ordering during normalization.

## Acceptance criteria

- Object-dtype categorical columns with safely normalizable repeated string/category values use the native factorizer without changing category labels, sorted palette order, or code assignments.
- Mixed-object and null-containing columns retain the existing fallback behavior and exact label semantics.
- Existing categorical color and palette parity tests remain unchanged and pass.
- Add a regression/performance-oriented test proving the eligible object path avoids the N-entry Python label materialization, using a spy or equivalent deterministic assertion rather than a timing threshold.
- Large object categorical channels do not regress memory usage or introduce an additional full-size copy when normalization is unnecessary.
- Document the fallback conditions so future changes do not broaden the fast path without parity coverage.

harsh21234i:
## Summary
Fixes #471
  - Add a safe native factorization path for homogeneous object-backed
    string categories.

  - Preserve existing fallback behavior for mixed objects, null-like values,
    bytes, and custom string semantics.

  - Keep the existing low-cardinality probe as the fast-path gate.
  - Preserve category ordering, codes, and count semantics.
  - Add regression coverage proving only unique labels cross the Python
    label path.

  ## Testing

  - tests/test_custom_ramps_and_palette.py: 92 passed, 1 skipped
  - tests/test_components.py: 103 passed
  - Ruff check and format passed
  - git diff --check passed

<a href="https://cubic.dev/pr/reflex-dev/xy/pull/476?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved categorical factorization for arrays containing only Python strings.
  * Ensured category order, unique labels, counts, and codes remain consistent with fixed-width string behavior.
  * Preserved distinct categories containing NUL characters.
  * Maintained safe fallback handling for mixed values, missing values, bytes, custom string-like values, and unsuitable near-unique arrays.
  * Improved performance by limiting category sampling during factorization.

* **Tests**
  * Added coverage for string-only, mixed-value, missing-value, NUL-containing, and near-unique category scenarios.

### DIFF
--- a/python/xy/channels.py
+++ b/python/xy/channels.py
@@ -371,6 +371,25 @@ def _use_native_fixed_factorizer(arr: np.ndarray) -> bool:
     return distinct < near_unique * len(probe)
 
 
+def _normalize_object_strings(arr: np.ndarray) -> Optional[np.ndarray]:
+    """Return object-backed strings as fixed-width Unicode when safe.
+
+    Object arrays can contain values with display-label semantics that cannot
+    be represented by a simple Unicode cast. Restrict this path to exact
+    Python strings so mixed objects, missing values, and custom ``__str__``
+    implementations retain the canonical fallback behavior.
+    """
+    if arr.dtype.kind != "O" or arr.size == 0:
+        return None
+    values = arr.reshape(-1)
+    width = 0
+    for value in values:
+        if type(value) is not str:
+            return None
+        width = max(width, len(value))
+    return np.asarray(arr, dtype=f"<U{max(width, 1)}")
+
+
 def _factorize_categories(
     arr: np.ndarray,
 ) -> tuple[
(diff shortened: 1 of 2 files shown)


harsh21234i:
Hi @Alek99 Could you please review the implementation and let me know if any changes are needed?

PR #489, opened 2026-08-13
Fix f32 precision loss for high-rate datetime axes (issue #487)

### LINKED ISSUE #487: Cannot render plots with high datetime accuracy - needs 64bit datetime support in the renderer?
Awesome tool - thanks for releasing it, it is very good.
I have been testing it out with some time series science data - a line chart of values with datetimes on the x axis and I think the renderer struggles if you have large amounts of datetime data of the order of hundreds of datapoints per second, plotting many hours of data, so millions of data items to plot

It ends up looking like this - the x axis datetimes are getting quantised i think

Image

And it should look more like this without the gaps/quantisation by forcing the data to a lower rate:

Image

My AI coding companion reckons this on the cause of the issue:

> looks like a bug in the xy charting library (v0.0.6): its WebGL client reconstructs each point's pixel position from a shared GLSL decode helper, xyDecode(encoded, meta) = encoded/meta.y + meta.x, entirely in 32-bit float. A real calendar-date epoch (~1.7e12 ms since 1970) exceeds  f32's ~2^24 integer precision budget, so every point in a narrow window rounds to the same pixel column. This affects every mark type (line/scatter/step/stairs/segments all funnel through the same shader)

alastairtree:
Aiming to fix https://github.com/reflex-dev/xy/issues/487 and as requested by @Alek99 am submitting thisPR which is just a copilot attempt at a fix and is not tested or reviewed properly, sorry! I am just trying to get the bug fixed but have very limited time to contribute properly.

Calendar timestamps (~1.7 × 10¹² ms since epoch) exceed f32's ~2²⁴ integer precision budget, causing the shader to collapse every timestamp within a ~200 s window to the same pixel column — producing the quantised, gapped rendering seen with high-rate time-series data.

<a href="https://cubic.dev/pr/reflex-dev/xy/pull/489?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved chart positioning and coordinate mapping for linear axes, especially with large values such as millisecond timestamps.
  - Increased numerical precision during pan and zoom to preserve accurate pixel spacing and endpoint placement.
  - Improved rendering consistency across bars, lines, areas, ribbons, funnels, meshes, rectangles, segments, and points.
  - Improved handling of zero, very small, degenerate, and invalid axis ranges.
  - Preserved coordinate transformation behavior for nonlinear axes, including logarithmic scales.
  - Improved bar edge placement when axis boundaries fall outside the valid domain.


js/src/40_gl.ts
```diff
@@ -129,6 +129,13 @@ float xyAxisCoord(float encoded, vec2 meta, int mode, float constant) {
   return value;
 }
 float xyMap(float encoded, vec2 map, vec2 meta, int mode, float constant) {
+  // For linear axes (mode 0) the CPU has already folded the column offset into
+  // the affine constants (map.x, map.y) in f64 (§4/§16). Apply them directly to
+  // the offset-encoded value to avoid reconstructing the large absolute coordinate
+  // in f32, which would discard low bits for high-magnitude axes (e.g. ms-since-
+  // epoch datetime).  Non-linear axes decode first because their transforms are
+  // not affine.
+  if (mode == 0) return encoded * map.x + map.y;
```
Alek99: [P1] Could we keep the folded map column-specific here? `mode == 0` now ignores `meta`, but the segment, ribbon, funnel, and mesh shaders still reuse a map built from `x0`/`y0` for other independently encoded columns. I verified this in the browser: a segment that spans 452 px on `main` disappears completely on this branch, and `render_smoke_nonumpy.py` flips from `funnel=1` to `funnel=0`. Per-column maps (or a shared encoding for same-axis columns) should fix this; the area perimeter's base map needs the same treatment.

js/src/50_chartview.ts
```diff
@@ -5398,6 +5398,18 @@ export class ChartView {
...
     const axis = this._axis(axisId);
+    // For linear axes the shader applies the map directly to encoded values (§4/§16):
+    //   encoded * mul + add  where  encoded = (v - offset) * scale
+    // Fold the column offset into the affine constants here in f64 so the shader
+    // never reconstructs the large absolute coordinate in f32.
+    if (this._axisMode(axisId) === 0) {
+      if (!Number.isFinite(hi - lo) || hi === lo) return [0, -2];
+      const scale = (meta && meta.scale) ? meta.scale : 1;
+      const offset = (meta && Number.isFinite(meta.offset)) ? meta.offset : 0;
+      const mul = 2 / ((hi - lo) * scale);
```
Alek99: [P2] One edge case here: `mul` is now measured per encoded unit, but `BAR_VS` still multiplies `u_pmap.x` by a width expressed in data units. With a legal encoding scale of `0.02` (`bar([-5e38, 5e38], width=1e38)`), the expected clip-space half-width changes from `0.0858` to `4.288`, so the bars cover the chart. Could we keep/pass a data-space slope for width, or compensate for the position scale?

alastairtree:
@Alek99 I've had another look at this and think it's ready for another review.

PR #494, opened 2026-08-14
fix(pyplot): stop minorticks_on() pinning automatic major ticks

### LINKED ISSUE #493: pyplot: `minorticks_on()` freezes the major tick positions, so labels vanish on zoom
Calling `minorticks_on()` turns the major ticks from an automatic axis into a fixed list of positions computed for the initial view. Zoom in interactively and the labels disappear once you get past roughly a factor of two, because the view no longer contains any of the frozen positions. Without `minorticks_on()` the labels rescale normally.

## Reproducer

```python
import xy.pyplot as plt


def x_axis(minor):
    fig, ax = plt.subplots(figsize=(6, 4))
    ax.plot([0, 10], [0, 10])
    if minor:
        ax.minorticks_on()
    ax.tick_params(which="major", direction="in", length=6,
                   bottom=True, top=True, left=True, right=True)
    payload = ax._build_chart(600, 400).figure().build_payload()
    spec = payload[0] if isinstance(payload, tuple) else payload
    return spec["x_axis"]


without, with_minor = x_axis(False), x_axis(True)
for key in ("tick_values", "tick_count", "minor_tick_values"):
    print(f"{key}")
    print(f"  without minorticks_on(): {str(without.get(key))[:60]}")
    print(f"  with    minorticks_on(): {str(with_minor.get(key))[:60]}")
```

On main (`84cb0e1`), Python 3.14, macOS:

```
tick_values
  without minorticks_on(): None
  with    minorticks_on(): [0.0, 2.0, 4.0, 6.0, 8.0, 10.0]
tick_count
  without minorticks_on(): 9
  with    minorticks_on(): 6
minor_tick_values
  without minorticks_on(): None
  with    minorticks_on(): [-0.5, 0.5, 1.0, 1.5, 2.5, 3.0, 3.5, 4.5, 5.0, 5.5
```

Without minor ticks the axis ships no authored major positions, so the client generates ticks for whatever view it is showing. With minor ticks the majors become six fixed values. The client then filters that fixed list to the visible window, so zooming into leaves nothing to label.

To see the visible effect, render the two figures in a notebook and zoom: the one with `minorticks_on()` loses its labels and keeps its tick marks.

## The problem

[`_apply_tickers`](https://github.com/reflex-dev/xy/blob/84cb0e1057d9a7e644b11fb687840074288a1746/python/xy/pyplot/_axes.py#L7192) returns early when nothing is registered:

```python
if locator is None and formatter is None and minor_locator is None and not is_log:
    return
```

[`minorticks_on()`](https://github.com/reflex-dev/xy/blob/84cb0e1057d9a7e644b11fb687840074288a1746/python/xy/pyplot/_axes.py#L5184) registers a `minor_locator`, so the guard is not triggered and execution reaches [`props["tick_values"] = ...`](https://github.com/reflex-dev/xy/blob/84cb0e1057d9a7e644b11fb687840074288a1746/python/xy/pyplot/_axes.py#L7248), which writes the majors as an authored list.

That write makes sense when the caller sets a major locator or formatter, since those define the positions. But, by registering a minor locator my major ticks shouldn't get pinned. Internally I would expect a redivision of the ticks for the minors, similar to `matplotlib`'s `AutoLocator`.

MoAly98:
Closes #493.

## The problem

`minorticks_on()` freezes the major tick positions. Asking for minor ticks turns an automatic axis into a fixed one, and because the renderer filters authored positions to the visible window, the tick labels disappear once you zoom past them. Without `minorticks_on()` the labels rescale normally.

Visible in the compiled payload, no browser needed:

```python
import xy.pyplot as plt

def x_axis(minor):
    fig, ax = plt.subplots(figsize=(6, 4))
    ax.plot([0, 10], [0, 10])
    if minor:
        ax.minorticks_on()
    spec, _blob = ax._build_chart(600, 400).figure().build_payload()
    return spec["x_axis"]

for key in ("tick_values", "tick_count"):
    print(key, x_axis(False).get(key), "->", x_axis(True).get(key))
```

Before this change:

```
tick_values None -> [0.0, 2.0, 4.0, 6.0, 8.0, 10.0]
tick_count  9    -> 6
```

## The cause

`_apply_tickers` returns early when nothing is registered:

```python
if locator is None and formatter is None and minor_locator is None and not is_log:
    return
```

`minorticks_on()` registers a `minor_locator`, so that guard stops short-circuiting and execution reaches `props["tick_values"] = ...`, which publishes the majors as authored positions.

That write is right when the caller sets a major locator or formatter, since those define where the majors go. It is wrong when the only thing registered is a minor locator: the majors were still chosen by `AutoLocator`, and pinning them is not correct.

## The change

Compute `ticks` as before. It is still needed to subdivide between the majors, but only publish `tick_values` and `tick_count` when the majors were actually determined by the caller:

```python
majors_are_auto = (
    locator is None
    and formatter is None
    and authored_labels is None
    and not is_log
    and "tick_values" not in props
)
```

After:

```
tick_values None -> None
tick_count  9    -> 9
```

with `minor_tick_values` still populated.

## Testing

- New regression test asserting `minorticks_on()` leaves `tick_values` unset while minor values are still emitted. It fails on `main` with `assert [0.0, 2.0, 4.0, 6.0, 8.0, 10.0] is None`.
- Confirmed by hand in a notebook: with minor ticks on, the tick labels now survive zooming instead of disappearing past roughly 2x.

## Caveat: this does not make minor ticks regenerate

The majors are automatic again, but the minors are not. `minor_tick_values` is the only minor field on the wire, and the client filters that fixed list to the view rather than generating anything:

```js
const minorTicks = (axis, axisId) => {
  if (!Array.isArray(axis.minor_tick_values)) return [];
  const [lo, hi] = this._axisRange(axisId);
  return axis.minor_tick_values.map(Number).filter((v) => v >= a && v <= b);
};
```

So the minor positions computed for the initial domain are all that ever exist. Zoom in far enough and none of them fall inside the view, leaving regenerating majors with no minors between them.

Fixing that properly means shipping the subdivision rule rather than the positions, something like `minor_subdivisions`,  and having the client subdivide between the majors it generates. That is a wire field plus client work, which would be a fair bit more work, and I'm not confident enough around the codebase  to tackle it. 

<a href="https://cubic.dev/pr/reflex-dev/xy/pull/494?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic major tick handling on linear axes.
  * Preserved explicit, formatted, logarithmic, and user-authored tick configurations.
  * Maintained correct tick labels and positions across shared nonlinear axes, including symlog, asinh, and logit scales.
  * Ensured minor ticks do not disrupt automatic major tick generation.

* **Tests**
  * Added regression coverage for automatic and nonlinear axis tick behavior.

### DIFF
--- a/python/xy/pyplot/_axes.py
+++ b/python/xy/pyplot/_axes.py
@@ -7245,10 +7245,6 @@ class Axes(PlotTypeMixin):
                 pad = (hi - lo) * 1e-9
                 ticks = ticks[(ticks >= lo - pad) & (ticks <= hi + pad)]
             auto_log = is_log
-        # A minor locator alone must not pin the majors. `ticks` is still needed
-        # below to subdivide between them, but authoring `tick_values` here would
-        # freeze an otherwise automatic axis: the renderer filters authored
-        # positions to the visible window, so zooming past them drops the labels.
         majors_are_auto = (
             locator is None
             and formatter is None
@@ -7256,8 +7252,6 @@ class Axes(PlotTypeMixin):
             and not is_log
             and "tick_values" not in props
         )
-        if not majors_are_auto:
-            props["tick_values"] = list(map(float, _scale_values(ticks, spec)))
         if formatter is not None:
             props["tick_labels"] = _formatter_tick_labels(
                 formatter,
@@ -7275,11 +7269,13 @@ class Axes(PlotTypeMixin):
             props["tick_labels"] = [f"{value:g}" for value in ticks]
         else:
             props.pop("tick_labels", None)
-        if not majors_are_auto:
-            if len(ticks):
-                props["tick_count"] = len(ticks)
-            else:
-                props.pop("tick_count", None)
+        # Authoring positions freezes an automatic axis; labels need them anyway.
+        if not majors_are_auto or "tick_labels" in props:
+            props["tick_values"] = list(map(float, _scale_values(ticks, spec)))
+        if len(ticks):
+            props["tick_count"] = len(ticks)
+        else:
+            props.pop("tick_count", None)
         if promoted_minor:
             props.pop("minor_tick_values", None)
             props["style"] = {
--- a/python/xy/pyplot/_axes.py
+++ b/python/xy/pyplot/_axes.py
@@ -7245,7 +7245,19 @@ class Axes(PlotTypeMixin):
                 pad = (hi - lo) * 1e-9
                 ticks = ticks[(ticks >= lo - pad) & (ticks <= hi + pad)]
             auto_log = is_log
-        props["tick_values"] = list(map(float, _scale_values(ticks, spec)))
+        # A minor locator alone must not pin the majors. `ticks` is still needed
+        # below to subdivide between them, but authoring `tick_values` here would
+        # freeze an otherwise automatic axis: the renderer filters authored
+        # positions to the visible window, so zooming past them drops the labels.
+        majors_are_auto = (
+            locator is None
+            and formatter is None
+            and authored_labels is None
+            and not is_log
+            and "tick_values" not in props
+        )
+        if not majors_are_auto:
+            props["tick_values"] = list(map(float, _scale_values(ticks, spec)))
         if formatter is not None:
             props["tick_labels"] = _formatter_tick_labels(
                 formatter,
@@ -7263,10 +7275,11 @@ class Axes(PlotTypeMixin):
             props["tick_labels"] = [f"{value:g}" for value in ticks]
         else:
             props.pop("tick_labels", None)
-        if len(ticks):
-            props["tick_count"] = len(ticks)
-        else:
-            props.pop("tick_count", None)
+        if not majors_are_auto:
+            if len(ticks):
+                props["tick_count"] = len(ticks)
+            else:
+                props.pop("tick_count", None)
         if promoted_minor:
             props.pop("minor_tick_values", None)
             props["style"] = {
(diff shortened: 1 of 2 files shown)


Alek99:
Hi thanks for the catches 

The two logit comments are pointing at the same issue, and I agree the test should use values inside (0, 1). Small nuance: the emitted tick positions are actually finite on the current head because it falls back to a default domain, but that still means the test isn’t exercising a real logit dataset.

The nonlinear label/position issue looks properly fixed in 8e7bff6.

I don’t think the locator_params(nbins=4) issue is fully resolved, though. minorticks_on() still changes the count from 4 to 6, so the ladders are synchronized by ignoring the requested count rather than using it to calculate the minors.

I also found a similar problem with datetime axes: the client generates calendar-based major ticks, while the minor ticks still come from the numeric-millisecond locator, so the two sets don’t line up.

Overall, the original disappearing-label bug is fixed, but I think the tick-count and datetime cases should be addressed before merging.

I can put the finishes on this pr or you can either way works, appreciate the contribution

PR #497, opened 2026-08-17
Fix state-backed docs module identity

Alek99:
## Summary

- derive executable Markdown module identities from each page's stable docs-relative path instead of its absolute checkout path
- use that identity consistently for fence cache invalidation and execution
- add a regression test that renders the same executable page from different frontend and backend build roots

## Root cause

Reflex includes the dynamically executed module name in each `State` subclass's full name. The docs frontend is compiled under a GitHub Actions checkout such as `/home/runner/work/xy/xy`, while the deployed backend runs under `/app`. Using the absolute Markdown source path therefore produced different State identifiers in the frontend and backend. The client could not dispatch the backend State update, so the state-backed XY data handle never hydrated and the chart remained blank.

## Impact

State-backed charts and other executable docs examples that define Reflex State now keep the same State identity across split frontend/backend builds.

## Validation

- `uv run --project docs/app --no-sync pytest docs/app/tests -q` — 120 passed, 2 skipped
- focused state-backed and split-build regression tests — 2 passed
- Ruff lint and format checks
- sitemap, Markdown asset, and generated HTML route validators
- production-mode docs build checked in a fresh browser tab: the state-backed chart rendered its plot area with no State-dispatch or unknown-token warning

<a href="https://cubic.dev/pr/reflex-dev/xy/pull/497?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed executable documentation examples so their identity remains stable across different build and checkout locations.
  - Ensured edited documentation pages correctly invalidate outdated code-example snapshots.
  - Improved consistency when rendering and executing code examples from documentation pages.

### DIFF
--- a/docs/app/tests/test_docs_site.py
+++ b/docs/app/tests/test_docs_site.py
@@ -66,6 +66,7 @@ from xy_docs.gallery import (
 )
 from xy_docs.markdown import (
     XyDocsMarkdownTransformer,
+    _page_virtual_filepath,
     _split_demo_data,
     page_with_api_reference_toc,
     render_xy_markdown_page,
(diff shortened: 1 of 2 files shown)


Alek99:
Production-mode validation (`reflex run --env prod --single-port`): the state-backed data preview hydrates and renders the chart. A fresh browser session reported no `no dispatch function` or `unknown figure token` warnings.

State-backed XY chart rendered after the fix

FarhanAliRaza (approved):

PR #501, opened 2026-08-25
Fix idle-chart blank/flicker: retain last frame through failed restore retries, recycle wedged contexts

FarhanAliRaza:
## Problem

Charts left idle for a while started disappearing and reappearing at random, and could stay blank until a full page refresh (reported against a deployed Reflex app; reproduced on demand in-browser).

Both symptoms are one recovery loop: context lost → restore → client rebuild → transient failure (Chromium's `shader compile: null` under process-wide GPU pressure) → backoff retry.

- **Flicker / long blanks:** `_initGl` runs on every retry and sized the canvas backing stores unconditionally — a canvas `width`/`height` write clears the canvas **even at the same value**, so every failed rebuild wiped the last presented frame. Blank for the whole failing stretch; brief reappearances when an attempt succeeded.
- **Refresh required:** a wedged driver can leave a context reporting itself live (`isContextLost()` false) while every rebuild against it fails. The loop retried the same poisoned context forever; only a reload allocated a fresh one.

## Fix

- `_initGl` resizes surfaces only when dimensions actually change → the last good frame survives every failed retry. Recovery is seamless-or-stale, never blank (`data-xy-ctx` still reports the state).
- New `GLHost.recycleSurface()`: after 3 consecutive rebuild failures the client retires the shared surface and rebuilds every client on a fresh context — the in-page equivalent of the reload, riding the existing loss/replace machinery and backoff. Counter resets on the first success, so ordinary transient pressure keeps cheap same-context retries.

## Evidence

- `test_failed_client_rebuild_keeps_last_presented_frame` — pixels retained on the presentation canvas while rebuilds fail continuously, recovery after they heal. **Fails on old code** (canvas blank), passes with the fix.
- `test_wedged_live_context_escalates_to_fresh_surface` — poisons exactly the current context object (live but failing every compile); chart must come back live on a replacement context without a reload. **Times out retrying forever on old code**, passes with the fix.
- Reproduced the reported blank state pixel-for-pixel on the deployed app by injecting the same failure; verified live that the fixed bundle holds its frame through sustained failures and returns to `live` automatically.
- Full `test_shared_glhost.py` + `test_context_loss_preserves_view.py` (12 passed), `render_smoke_nonumpy`, `append_stream_smoke`, `node js/build.mjs`, pre-commit + ruff all green.

## Spec

Dossier §18 now records both contracts: surface backing stores resize only on real dimension changes (recovery never blanks), and retries must be as powerful as a page reload (escalation to a fresh surface).

<a href="https://cubic.dev/pr/reflex-dev/xy/pull/501?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved graphics recovery after temporary WebGL or shader failures.
  * Retries transient recovery failures with controlled backoff.
  * Preserves the last rendered frame while recovery is in progress.
  * Refreshes the graphics surface after three consecutive recovery failures.
  * Prevents unnecessary canvas resizing during transient rendering issues.
  * Resets recovery tracking after successful restoration or genuine context loss.

* **Tests**
  * Added coverage for frame preservation and stalled graphics-context recovery.

### DIFF
--- a/js/src/50_chartview.ts
+++ b/js/src/50_chartview.ts
@@ -681,6 +681,7 @@ export class ChartView {
     this._governorRegistered = false;
     this._glHostRecoveryTimer = null;
     this._glHostRecoveryDelay = 0;
+    this._glHostRebuildFailures = 0;
     if (this._ctxVisible) this._ctxSeenSeq = XY_CONTEXT_GOVERNOR.seq++;
     this._contextLossCount = 0;
     this._contextRestoreCount = 0;
@@ -2047,7 +2048,23 @@ export class ChartView {
             if (this.gl && !this.gl.isContextLost()) {
               try { this._destroyGlResources(); } catch (_cleanupError) {}
             }
-            this._scheduleGlHostClientRecovery();
+            // A host context can stay "live" (`isContextLost()` false) while a
+            // wedged driver fails every rebuild against it. Retrying on that
+            // same context can then never succeed — a state users previously
+            // escaped only by reloading the page. After three consecutive
+            // failures, recycle the shared surface instead: a fresh context is
+            // exactly what a reload buys. (The capability check keeps a v1
+            // host from an older duplicate bundle on plain retries.)
+            this._glHostRebuildFailures = (this._glHostRebuildFailures || 0) + 1;
+            if (
+              this._glHostRebuildFailures >= 3 &&
+              typeof this._glHost.recycleSurface === "function"
+            ) {
+              this._glHostRebuildFailures = 0;
+              this._glHost.recycleSurface(); // host-wide restored event drives the next attempt
+            } else {
+              this._scheduleGlHostClientRecovery();
+            }
           } else {
             this._scheduleContextRecovery();
           }
@@ -2069,6 +2086,7 @@ export class ChartView {
       }
       this._contextRestoreCount += 1;
       this._contextRecoveryError = null;
+      this._glHostRebuildFailures = 0;
       this._ctxRecoveryDelay = 0;
       clearTimeout(this._glHostRecoveryTimer);
       this._glHostRecoveryTimer = null;
@@ -4358,14 +4376,22 @@ export class ChartView {
   _initGl(buffer) {
     const dpr = window.devicePixelRatio || 1;
     this.dpr = dpr;
-    this.canvas.width = this.plot.w * dpr;
-    this.canvas.height = this.plot.h * dpr;
-    this.chrome.width = this.size.w * dpr;
-    this.chrome.height = this.size.h * dpr;
+    // A canvas backing-store write clears the canvas even when the value is
+    // unchanged, and _initGl also runs on every context-restore retry. The
+    // last presented frame must survive a rebuild that fails transiently
+    // under GPU pressure (§18) — wiping it here turned each failed retry
+    // into a visible blank/flicker until a retry finally succeeded.
+    const sizeSurface = (surface, w, h) => {
+      w = Math.trunc(w);
+      h = Math.trunc(h);
+      if (surface.width !== w) surface.width = w;
+      if (surface.height !== h) surface.height = h;
+    };
+    sizeSurface(this.canvas, this.plot.w * dpr, this.plot.h * dpr);
+    sizeSurface(this.chrome, this.size.w * dpr, this.size.h * dpr);
     this.chrome.style.width = this.size.w + "px";
     this.chrome.style.height = this.size.h + "px";
-    this.overlay.width = this.size.w * dpr;
-    this.overlay.height = this.size.h * dpr;
+    sizeSurface(this.overlay, this.size.w * dpr, this.size.h * dpr);
     this.overlay.style.width = this.size.w + "px";
     this.overlay.style.height = this.size.h + "px";
 
--- a/js/src/50_chartview.ts
+++ b/js/src/50_chartview.ts
@@ -1916,11 +1916,9 @@ export class ChartView {
       if (this._governorRegistered) XY_CONTEXT_GOVERNOR._announceLive();
       this._contextLossCount += 1;
       this._contextRecoveryError = null;
-      // A genuine loss starts a fresh recovery cycle (and already delivers
-      // the fresh-context remedy escalation exists for), so rebuild failures
-      // from a previous cycle are not evidence about the next context —
-      // carrying them over would recycle the whole host after one post-loss
-      // failure instead of three consecutive ones.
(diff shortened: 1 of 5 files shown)


js/src/50_chartview.ts
```diff
@@ -2047,7 +2048,23 @@ export class ChartView {
...
+            // failures, recycle the shared surface instead: a fresh context is
+            // exactly what a reload buys. (The capability check keeps a v1
+            // host from an older duplicate bundle on plain retries.)
+            this._glHostRebuildFailures = (this._glHostRebuildFailures || 0) + 1;
+            if (
+              this._glHostRebuildFailures >= 3 &&
+              typeof this._glHost.recycleSurface === "function"
+            ) {
+              this._glHostRebuildFailures = 0;
+              this._glHost.recycleSurface(); // host-wide restored event drives the next attempt
```
masenf, line 2070: what happens if this fails after resetting the _glHostRebuildFailures back to 0?

just an idea, but should we try to recycleSurface() first, catch any errors, and reset the count after it's clean that way we can surface the errors in the console and avoid a constant recycle loop?

js/src/50_chartview.ts
```diff
@@ -681,6 +681,7 @@ export class ChartView {
     this._governorRegistered = false;
     this._glHostRecoveryTimer = null;
     this._glHostRecoveryDelay = 0;
+    this._glHostRebuildFailures = 0;
```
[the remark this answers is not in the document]
FarhanAliRaza: Valid — fixed in 04d5a4dd. The counter now resets in the webglcontextlost handler (alongside `_contextLossCount += 1`), so failures from a previous cycle never carry across a genuine loss; escalation still requires 3 consecutive failures within one restored-context cycle, and the wedged path is unaffected (no loss event fires between its retries). Regression test: `test_genuine_loss_resets_rebuild_failure_count` — verified red before the fix (stale count of 2 + one post-loss failure escalated and replaced the context) and green after (same-context retry recovers, generation delta exactly 1).

js/src/50_chartview.ts
```diff
@@ -4358,14 +4376,22 @@ export class ChartView {
...
     this.dpr = dpr;
-    this.canvas.width = this.plot.w * dpr;
-    this.canvas.height = this.plot.h * dpr;
-    this.chrome.width = this.size.w * dpr;
-    this.chrome.height = this.size.h * dpr;
+    // A canvas backing-store write clears the canvas even when the value is
+    // unchanged, and _initGl also runs on every context-restore retry. The
+    // last presented frame must survive a rebuild that fails transiently
+    // under GPU pressure (§18) — wiping it here turned each failed retry
+    // into a visible blank/flicker until a retry finally succeeded.
```
masenf: get rid of the narrative of _how_ we got here, just keep the comment explaining why this is needed

js/src/50_chartview.ts
FarhanAliRaza: it will not be cheap it tries cheap solution first beacuse that works in most of cases.

js/src/50_chartview.ts
FarhanAliRaza: we should add this to Claude.md in repo, in both reflex and here ?

FarhanAliRaza: Trimmed in a3a1ad5f — all four comments this PR added (the _initGl resize guard, the escalation site, the loss-handler reset, and recycleSurface's doc) now state only the invariant; the defect story stays in the fix commits and the PR description. The two test docstrings got the same treatment. Keeping the CLAUDE.md guideline out of this PR — can land separately here and in reflex.

js/src/50_chartview.ts
FarhanAliRaza: recycleSurface() itself can't fail — it only marks the host lost and schedules recovery (no GL calls; the client-lost fan-out already try/catches per client). The part that can fail is the later surface replacement, and that already has the properties you're describing: _replaceLostSurface runs in a try/catch, logs 'xy: shared WebGL host replacement failed' to the console, and re-arms with 250ms→4s exponential backoff — it never throws back to the caller. A constant recycle loop can't form: each escalation additionally requires 3 fresh client rebuild failures (50/100/200ms backoff) on the *new* context, since the recycle's own loss event resets the counter (04d5a4dd). So 'reset after it's clean' effectively already happens — the authoritative reset is on success in the restored handler and on each loss; the pre-call reset at the escalation site is just the synchronous belt. Worst case under permanent failure: one replacement attempt every ~4s with a console error each time, charts holding their last frame.

js/src/50_chartview.ts
```diff
@@ -1915,6 +1916,10 @@ export class ChartView {
       if (this._governorRegistered) XY_CONTEXT_GOVERNOR._announceLive();
       this._contextLossCount += 1;
       this._contextRecoveryError = null;
+      // A genuine loss starts a fresh recovery cycle with a fresh context, so
+      // prior rebuild failures are no evidence about it — escalation counts
+      // consecutive failures within one cycle only.
+      this._glHostRebuildFailures = 0;
```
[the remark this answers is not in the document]
FarhanAliRaza: Valid — fixed in 03b51a55. The duplicate-loss guard did return before the counter reset, so a genuine host loss landing on a client still lost from a failed rebuild kept the stale count, and one transient failure on the fresh context could then escalate a host-wide recycleSurface(). Since GLHost fans out losses only on a live→lost transition, any loss reaching the handler heralds a fresh context — the reset now runs before the guard (duplicate-loss telemetry suppression unchanged). Regression test: test_loss_while_client_lost_resets_rebuild_failure_count (failed before the fix with sameContext=false, generationDelta=2).

js/src/50_chartview.ts
FarhanAliRaza: Follow-up: cubic found one gap in the claim above — the loss-event counter reset was skipped for a client already lost from a failed rebuild (the duplicate-loss guard returned first), so a peer client mid-retry during a recycle could keep a stale count and escalate again off a single failure on the new context. Fixed in 03b51a55 by resetting before the guard, with a regression test (test_loss_while_client_lost_resets_rebuild_failure_count). The no-recycle-loop guarantee now holds unconditionally: every escalation requires 3 fresh failures on the current context.

masenf (approved):

